Issue #89, opened 2024-03-13
Construction started after technology ended. Should be unrelated 


felixvonsamson:
![Screenshot_20240313-212917.png](https://github.com/felixvonsamson/Energetica/assets/55402056/f957c3dc-4c92-4b4d-941e-3e80e0c08939)

PR #96, opened 2024-03-18
bug fixes

felixvonsamson:
- resource purchase bug fixed 
- finished technology research will not trigger the start of a new construction (bug)
- order of facilities in storage graph fixed


mglst:
Fixes #89

PR #207, opened 2024-11-22
Fix: Text formatting for lvl 0 functional facility upgrades

### LINKED ISSUE #190: Fix text formatting for pairs (->) when the first value is 0
This should say 15MW rather than 15'000'000W
image

mglst:
Fixes #190


felixvonsamson (approved):
I don't see why none is better than 0 but ok

PR #215, opened 2024-12-04
Feat landing page

mglst:
Add a landing page.

For now a word-for-word copy of the recent poster.


mglst:
@felixvonsamson could you replace the images in `/static/images/landing`? For now these are screenshots

PR #247, opened 2025-01-07
Feature/fix randomness

mglst:


energetica/database/engine_data.py
```diff
@@ -54,10 +56,22 @@ def add_bid(self, bid_name: str) -> None:
             "lithium_ion_batteries": 940.0,
             "solid_state_batteries": 900.0,
         }[bid_name]
+        # seed based off engine seed, ask/bid, bid name, and player username
+        seed_hash = hash(
+            (
+                engine.data["random_seed"],
+                "bid",
+                bid_name,
+                player.username,
```
yassir-akram, line 65: username not unique, use user id

energetica/database/engine_data.py
```diff
@@ -54,10 +56,22 @@ def add_bid(self, bid_name: str) -> None:
...
+            )
+        )
+        random.seed(seed_hash)
+        added_randomness = random.uniform(-15, 15)
+        self.bid_prices[bid_name] = default_bid_price + added_randomness
```
yassir-akram, line 70: why is there randomness?

energetica/utils/assets.py
```diff
@@ -107,10 +107,20 @@ def finish_project(project: OngoingProject, *, skip_notifications: bool = False)
...
+        # facilities of that type the player has built. This ensures that the facility's random position is generated
+        # deterministically.
         if player.tile is None:
             raise GameError("Player has no tile")
+        seed_hash = hash(
+            (
+                engine.data["random_seed"],
+                player.tile.coordinates,
+                project.name,
+                ActiveFacility.count_when(name=project.name, player=player),
```
yassir-akram, line 120: why not just the id?

yassir-akram (changes requested):

energetica/database/engine_data.py
mglst: Discussed with Felix, to introduce more interesting prices on networks given that many players sont modify their default prices

energetica/database/engine_data.py
mglst: I'm reticent to use ID's, as it makes the determinism of the simulation dependent on ID attribution strategy

energetica/database/engine_data.py
yassir-akram: maybe a comment? or maybe I just didnt read

energetica/database/engine_data.py
yassir-akram: for user, we could have unique usernames, that would fix it, for facilities, it's trickier, it's trickier

energetica/database/engine_data.py
mglst: There was a todo, which was removed. I'll add a comment

PR #248, opened 2025-01-16
refactor: add randomization to starer network prices

mglst:


yassir-akram (commented):
overall I think create_bid_entry and create_buy_entry add too much complexity for nothing. I would just init the whole array at the creating of the player. You can decide

energetica/database/engine_data.py
```diff
@@ -37,9 +37,8 @@ class NetworkPrices:
 
     # Prices are randomized so that each player's prices are slightly different. This leads to more interesting
     # market dynamics, while still having reasonable default prices.
-    # TODO (mglst): add randomess to theses prices
-    bid_prices: dict[str, float] = field(default_factory=lambda: {"steam_engine": 125.0})
-    ask_prices: dict[str, float] = field(default_factory=lambda: {"industry": 1000.0, "construction": 1020.0})
+    bid_prices: dict[str, float] = field(default_factory=lambda: {})
+    ask_prices: dict[str, float] = field(default_factory=lambda: {})
```
felixvonsamson, line 41: isnt feied(dict) sufficient ? shouldn't we use defaultdicts ?

felixvonsamson (approved):

energetica/database/engine_data.py
mglst: The intention is indeed to make use of standard `dict`s rather than `defaultdict`s. There are some places in `production_update` however where I think we should take another look, and I'm not sure that the existing tests cover all cases

mglst:
> overall I think create_bid_entry and create_buy_entry add too much complexity for nothing. I would just init the whole array at the creating of the player. You can decide

This has been revised. All prices now exist upon initialisation. The randomness is added in the post_init method of player

felixvonsamson (approved):

energetica/database/player.py
```diff
@@ -129,6 +129,11 @@ class Player(DBModel, UserMixin):
     )
     socketio_clients: list = field(default_factory=list)
 
+    def __post_init__(self) -> None:
```
yassir-akram, line 132: why not in networkprices post_init?

energetica/database/engine_data.py
```diff
@@ -109,19 +84,22 @@ def create_ask_entry(self, ask_name: Ask, player: Player) -> None:
...
-                player.id,
-            )
-        )
-        random.seed(seed_hash)
-        added_randomness = random.uniform(-15, 15)
-        self.ask_prices[ask_name] = default_ask_price + added_randomness
+        }
+    )
+
+    def init_prices_with_randomness(self, player: Player) -> None:
```
yassir-akram, line 90: instead of having a seed per bid price, I'd just create a vector random.uniform(-15, 15, len(self.bid_prices)) but what you did work

energetica/database/player.py
mglst: The seed uses the player id, so we need a reference to the player object to initialise the random prices. So we'd need such a `__post_init__` method in `NetworkPrices` to have the player as an argument - which seems to be possible using maybe `dataclass`'s `InitVar`, but I decided against looking further into it and ended up writing what we have right now

energetica/database/engine_data.py
mglst: One possible downside I see to this is that the randomness then depends on the order in which items appear. In the case we add or remove entries (which might happen if we merge the two nuclear power plants, and the small and large water dams) then the random prices would end up being different after the change, which might lead to desync issues when simulating action logs

yassir-akram: I know, If we remove stuff, then action_logger is just desync, if we add, I'd assume most of the time the behaviour of the game is changed anyway. Whatever solution is fine

energetica/database/player.py
yassir-akram: InitVar won't solve the problem, but got it

PR #249, opened 2025-01-19
Feature/strenum

mglst:


energetica/enums.py
```diff
@@ -39,9 +39,144 @@ class Renewable(StrEnum):
...
+    MINERAL_EXTRACTION = "mineral_extraction"
+    TRANSPORT_TECHNOLOGY = "transport_technology"
+    MATERIALS = "materials"
+    CIVIL_ENGINEERING = "civil_engineering"
+    AERODYNAMICS = "aerodynamics"
+    CHEMISTRY = "chemistry"
+    NUCLEAR_ENGINEERING = "nuclear_engineering"
+
+
+class SpecialAskType(StrEnum):
```
yassir-akram, line 130: Why Special? I find the name quite conterintuitive.

mglst: This was previously called `special_power_demand`. It shouldn't be called just `AskType` since this class is not exhaustive - any storage facility for example is a valid ask type. Indeed, there is are two type aliases - `Ask` and `Bid` defined in `engine_data.py` (arguably these could be renamed `AskType` and `BidType`) `Ask` includes but is not limited to `SpecialAskType`. I've updated the DocString slightly. Do you have a suggestion for a different name?

tests/unit/test_project_type.py
```diff
@@ -0,0 +1,77 @@
...
+    SpecialAskType,
+    StorageFacilityType,
+    TechnologyType,
+    WindFacilityType,
+    str_to_project_type,
+    str_to_project_type_extended,
+)
+
+
+def test_str_to_project_type():
```
yassir-akram, line 16: don't see the reason why there is a test for that

energetica/enums.py
yassir-akram: I think I'd just be more verbose: what is an ask, why special. Maybe RemainingAskType, AuxiliaryAskType, AuxiliaryElectricityConsumer je sais pas trop

mglst: What do you think of `NonFacilityAskType`?

tests/unit/test_project_type.py
mglst: I removed this test

PR #252, opened 2025-01-26
Refactor/strenums continued

mglst:


felixvonsamson (approved):

felixvonsamson:
Should we also use sternum for const_config ?

mglst:
I see you've noted that I've purposely avoided making modifications to const_config. I'm not really sure what would be best - we've discussed storing game config data in separate files, maybe CSV's - until then I've simply chosen not to change them

felixvonsamson:
Makes sense

yassir-akram (approved):

PR #255, opened 2025-01-28
Refactor/project multipliers

mglst:
Remove multiplier_1, multiplier_2, multiplier_3

### DIFF
--- a/energetica/utils/misc.py
+++ b/energetica/utils/misc.py
@@ -10,6 +10,7 @@ from flask import flash
 from noise import pnoise3
 from scipy.stats import norm
 
+from energetica import technology_effects
 from energetica.config.assets import river_discharge_seasonal
 from energetica.database.active_facility import ActiveFacility
 from energetica.database.map import HexTile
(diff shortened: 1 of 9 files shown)


energetica/utils/misc.py
```diff
@@ -184,12 +185,7 @@ def initialize_player(player: Player) -> None:
         position=(pos_x, pos_y),
         end_of_life=eol,
         player=player,
-        multipliers={
-            "price_multiplier": 1.0,
-            "multiplier_1": 1.0,
-            "multiplier_2": 1.0,
-            "multiplier_3": 1.0,
-        },
+        multipliers=technology_effects.current_multipliers(player, ControllableFacilityType.STEAM_ENGINE),
```
felixvonsamson, line 188: Is that needed ? Why is the post init not enough?

felixvonsamson (approved):

energetica/utils/misc.py
mglst: `OngoingProject`'s `__post_init__` method does indeed take care of correctly setting up the multipliers. However, here's we're calling the constructor for the `ActiveFacility` class, which doesn't have a `__post_init__` method, so we need to give the multipliers explicitly. There are two instances in the code base where we call the `ActiveFacility` constructor: in the `finish_project` method (where we get the multipliers from the `OngoingProject` that finished) and here, in the `initialize_player` function, hence why it's a bit of a special edge case and why we need to compute the multipliers

yassir-akram: this doesn't make sense to me, since this will be overwritten anyway

yassir-akram (commented):
Generally speaking, makes sense to me. However, it's quite hard to review everything in detail. So double check :)

energetica/utils/misc.py
yassir-akram: oh okk

yassir-akram (approved):

PR #257, opened 2025-01-30
Fix/climate events randomness

mglst:


mglst:
Note `check_climate_events` is the main entry point, so placing the seed at the very start should cover everythin

felixvonsamson (approved):

yassir-akram:
generally speaking, I think we should write code like this:
```
import numpy as np
rng = np.random.default_rng(seed=42)
arr2 = rng.random((3, 3))
```
Reasing being, if some other code uses anything random in parallel, we are fucked. Now of course this is not the case, but who knows in the future.

PR #259, opened 2025-03-31
feat: policies

mglst:


felixvonsamson (commented):
The file `test_policies.py` is an integration test to test some standard policies and the policy_runner.py is for experimentation ? is that correct ? If not, can you explain the distinction.

PR #260, opened 2025-04-12
Fix/237

mglst:


energetica/utils/assets.py
```diff
@@ -146,9 +146,14 @@ def finish_project(project: OngoingProject, *, skip_notifications: bool = False)
 
     if isinstance(project.project_type, FunctionalFacilityType):
         player.invalidate_recompute_and_dispatch_data_for_pages(
+            # Update power page if project is a warehouse, for the Combined Cycle facility
```
felixvonsamson, line 149: It's not only the Combined Cycle but all facilities that require fuel

felixvonsamson (commented):
It's not only the Combined Cycle but all facilities that require fuel

energetica/utils/assets.py
yassir-akram: more context please. Why update power_factilities for warehouse?

mglst: Warehouse lvl 2 is a requirement for Combined Cycle. So if the `project` which just finished is a warehouse, then we want to update the power facilities page to make sure that it shows as unlocked, if indeed it has been unlocked. But if it's industry, no need to check. Is the idea.

felixvonsamson (approved):

PR #263, opened 2025-04-15
Fix/240

### LINKED ISSUE #240: no toast to say not enough money when upgrading a facility with not enough money

mglst:
Fixes #240


felixvonsamson:
We should be consistend with the format of the error strings

PR #265, opened 2025-04-20
Refactor climate event datastructure

mglst:
Enhance the handling of climate events by utilizing a new ClimateEvent class for better organization and readability.


mglst:
Agreed, I don't feel great about the description function either. And I don't have good ideas about how to do it better. Perhaps the f strings could actually just be passed to the constructor of the new class, I haven't checked to see if that works actually

PR #272, opened 2025-06-17
API main merge

mglst:
These are the planned changes for API. This pull request is so that you can review and provide feedback

### DIFF
--- a/energetica/production_update.py
+++ b/energetica/production_update.py
@@ -118,7 +118,7 @@ def set_facilities_usage(new_values: dict, player: Player) -> None:
             else:
                 usage = new_values["storage"][storage_facility] / player.capacities[storage_facility]["capacity"]
             for af in ActiveFacility.filter_by(player=player, facility_type=storage_facility):
-                af.usage = usage
+                af.usage = usage  # type: ignore
 
     for extraction_facility in ExtractionFacilityType:
         if extraction_facility in player.capacities:
@@ -411,10 +411,10 @@ def calculate_generation_with_market(new_values: dict, market: dict, player: Pla
             max_prod = calculate_prod(
                 "max",
                 player,
-                facility,
+                facility,  # type: ignore
                 resource_reservations,
             )
-            price = player.network_prices.ask_prices[facility]
+            price = player.network_prices.ask_prices[facility]  # type: ignore
             capacity = max_prod - generation[facility]
             market = place_bid(market, player.id, capacity, price, facility)
 
@@ -461,7 +461,7 @@ def market_logic(new_values: dict, market: dict) -> None:
         demand["exports"] += quantity
         player.money += quantity * market_price / 3600 * engine.in_game_seconds_per_tick / 1_000_000
         revenue["exports"] += quantity * market_price / 3600 * engine.in_game_seconds_per_tick / 1_000_000
-        add_to_market_data(player.id, quantity, row.facility, export=True)
+        add_to_market_data(player.id, quantity, row.facility, export=True)  # type: ignore
 
     def buy(row: Any, market_price: Any, quantity: float | None = None) -> None:
         """Buy demanded power capacity."""
@@ -474,7 +474,7 @@ def market_logic(new_values: dict, market: dict) -> None:
         generation["imports"] += quantity
         player.money -= quantity * market_price / 3600 * engine.in_game_seconds_per_tick / 1_000_000
         revenue["imports"] -= quantity * market_price / 3600 * engine.in_game_seconds_per_tick / 1_000_000
-        add_to_market_data(player.id, quantity, row.facility, export=False)
+        add_to_market_data(player.id, quantity, row.facility, export=False)  # type: ignore
 
     market["player_exports"] = {}
     market["player_imports"] = {}
@@ -557,15 +557,15 @@ def market_optimum(offers_og: Any, demands_og: Any) -> tuple[float, float]:
     merged_table = merged_table.sort_values(by="cumul_capacities")
 
     for row in merged_table.itertuples():
-        if np.isnan(row.index_offer):
+        if np.isnan(row.index_offer):  # type: ignore
             price_d = row.price
         else:
             price_o = row.price
-        if price_d < price_o:
+        if price_d < price_o:  # type: ignore
             price = price_d
-            if np.isnan(row.index_offer):
+            if np.isnan(row.index_offer):  # type: ignore
                 price = price_o
-            return price, row.cumul_capacities
+            return price, row.cumul_capacities  # type: ignore
     raise ValueError("No market optimum found.")
 
 
--- a/energetica/routers/__init__.py
+++ b/energetica/routers/__init__.py
@@ -1,5 +1,6 @@
 import json
 import logging
+import urllib.parse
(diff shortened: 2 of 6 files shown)


energetica/database/messages.py
```diff
@@ -15,20 +13,21 @@
 
 
 @dataclass
-class Message(DBModel):
+class Message:
```
felixvonsamson, line 16: why did you remove messages fron the DBModel ?

energetica/database/player.py
```diff
@@ -57,7 +58,7 @@ class Player(DBModel):
 
     # inactive: bool = False  # True if account is inactive
     show_chat_disclaimer: bool = True
-    last_opened_chat: int = 0
+    last_opened_chat_id: int | None = None
```
felixvonsamson, line 61: The 0 (id of the general chat) was on purpose so that the default chat that opens at the start is the general chat

energetica/database/messages.py
```diff
@@ -40,10 +39,18 @@ class Chat(DBModel):
     name: str | None
     participants: set[Player]
     messages: list[Message] = field(default_factory=list)
-
-    last_read_message: dict[int, int | None] = field(
-        default_factory=lambda: defaultdict(partial(int, -1)),
-    )  # {player: message index in messages}
+    player_last_read_index: dict[int, int] = field(default_factory=dict)
```
felixvonsamson, line 42: We have to decide if we use -1 or None for the last_read_index of chats that have never been opened

energetica/database/messages.py
mglst: This is a subtle and important point, thanks for brining it up. The question is around the *location* of a resource. One location should of course only point to one resource - there shouldn't be ambiguity. As an analogy, my postcode is not a precise location of where I live. This is straightforward enough. On the other hand, locations should also be unique. That is, there should only be one way of specifying the location of a resource. If would be very inconvenient if there were two competing systems. As it turns out, this is the case for locations of Messages as it stands. By making Message inherit from DBModel, it automatically gets an id, which is a first location. However, messages always belong to a Chat. Chat's store a list - indexed - of its Message's, and they themselves inherit from DBModel. As such, the second way of specifying the location of a message is with the id of the Chat it belongs to, and the index of the message in that chat. By making DBModel no longer inherit from DBModel, this deduplicated and resolve this

energetica/database/messages.py
mglst: The option of not storing a key for the player comes down to the same thing as storing None when using `player_last_read_index.get(player.id)` since `get` returns `None` for missing keys.
This is what I suggest we go with

energetica/utils/network_helpers.py
```diff
@@ -19,33 +19,27 @@ def join_network(player: Player, network: Network | None) -> Network:
...
-    # websocket.rest_notify_network_change()
     return new_network
 
 
-# TODO (Felix): Move this to a method in Player
-# TODO(mglst): I'm not sure I agree. Why should this not be a method of Network for example?
-def leave_network(player: Player) -> None:
-    """Shared API method for a player to leave a network. Always succeeds."""
+def leave_network(player: Player) -> Network | None:
+    """Shared API method for a player to leave a network."""
```
felixvonsamson, line 42: we should remove the "shared API" it might be confusing now

energetica/utils/resource_market.py
```diff
@@ -11,32 +11,39 @@
 from energetica.utils.formatting import display_money, format_mass
 
 
-def put_resource_on_market(player: Player, fuel: Fuel, quantity: float, price: float) -> None:
+def put_resource_on_market(player: Player, fuel: Fuel, quantity: float, unit_price: float) -> ResourceOnSale:
```
felixvonsamson, line 14: create_ask_on_resource_market

energetica/utils/resource_market.py
```diff
@@ -11,32 +11,39 @@
...
         quantity=quantity,
-        price=price,
+        unit_price=unit_price,
         player=player,
     )
 
 
-def buy_resource_from_market(player: Player, quantity: float, sale: ResourceOnSale) -> None:
-    """Buy an offer from the resource market."""
+def buy_resource_from_market(player: Player, quantity: float, sale: ResourceOnSale) -> ResourceOnSale | None:
```
felixvonsamson, line 27: purchase_resource_from_market or just purchase_resource

felixvonsamson (changes requested):
I reviewed only the python. Added a lot of comments, some questions, some details, some actually important.

energetica/database/messages.py
mglst: @felixvonsamson if you're happy with this, go ahead and resolve this conversation

energetica/database/messages.py
mglst: Upon further reflection, why is this data something we want to store in the backend? Why would this not be stored in the browser?

energetica/database/network.py
```diff
@@ -49,3 +50,6 @@ def delete(self) -> None:
         network_path = f"instance/data/networks/{self.id}"
         shutil.rmtree(network_path)
         super().delete()
+
+    def to_schema(self) -> NetworkOut:
```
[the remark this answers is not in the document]
mglst: I've added to next meeting's minutes

energetica/routers/resource_market.py
```diff
@@ -0,0 +1,85 @@
...
+        quantity=request_data.quantity,
+        sale=sale,
+    )
+    if new_sale is None:
+        return None
+    return new_sale.to_schema()
+
+
+@router.patch("/asks/{ask_id}")
+async def patch_resource_market_ask(
```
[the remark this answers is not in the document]
mglst: Yes. Instead of a player buying their own ask in whole to delete it they should DELETE it, and instead of buying their own ask in part to modify quantities, they should PATCH it. Also they can now modify the price. That being said they shouldn't be able to change the resource type

energetica/schemas/chats.py
```diff
@@ -0,0 +1,60 @@
...
+class MessageListResponse(BaseModel):
+    """Response model for the message list."""
+
+    messages: list[MessageOut]
+
+
+class NewMessageRequest(BaseModel):
+    """Request model for sending a new message."""
+
+    new_message: str = Field(min_length=1, max_length=500)
```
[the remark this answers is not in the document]
mglst: I've doubled it

energetica/database/messages.py
felixvonsamson: Ok but the goal of the DBModel Class is to have all the data of the game stored in the same place (in engine.db_model_instances) and now if I understand it correctly, the content of the messages will not be saved there anymore.

energetica/database/messages.py
felixvonsamson: Because I want my messages that I read on my phone to also be marked as read on my computer (like whatsapp)
None is good

mglst (commented):
I suggest we close this merge request until further notice. Instead,

- All further changes should happen in sub-branches that should be merged back into this one
- Once all changes are done and we're ready to merge back into dev, a new merge request will be created and we'll review everything then

PR #273, opened 2025-06-18
Refactor/api weather

mglst:
Proposed changes for the API for weather

### DIFF
--- a/energetica/static/dashboard.js
+++ b/energetica/static/dashboard.js
@@ -5,9 +5,23 @@ function update_weather_conditions() {
     fetch("/api/v1/weather")
         .then((response) => response.json())
         .then((weather_data) => {
+            const month_name = [
+                "January",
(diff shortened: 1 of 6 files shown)


energetica/static/dashboard.js
felixvonsamson, line 1: I guess the months list could be used in different places in the frontend and should probably be in a separate file ? But I guess we can also do that when we switch to react.

felixvonsamson (approved):

energetica/static/dashboard.js
mglst: Agreed. We could store these in a statically served JSON file. That would make support for other languages easier, by just swapping out the JSON file. Then again there are a whole lot of places where we'd need to remove hard coded language. Maybe we should note that this is something we want to consider when planning the front end rewrite. Until then, let's remove language from the backend

PR #276, opened 2025-06-20
New policies functionality

mglst:
With these changes, it is now possibly to run the game with some players controlled by policies. These are configured with the `players.txt` file


mglst:
This merge request has a small amount of new functionality in the form of loading from the 'players.txt' file, which is an interesting new functionality, but it has not been satisfactorily implemented.

Largely, there is an issue with specification. It's not clear exactly what we want from these new policies specifically.

As such, the next step is to decide on a spec for what we want from policies. This should happen in an issue discussion. This would then unblock work on this branch.

PR #277, opened 2025-06-23
New API for projects

mglst:
Endpoints changes:

- `get_constructions` → `get("")`
- `request_queue_project` → `post("")`
- `request_cancel_project` → `delete("/{project_id}")`
- `request_toggle_pause_project` → `post("/{project_id}:pause")` & `post("/{project_id}:resume")`
- `request_decrease_project_priority` → `post("/{project_id}:decrease-priority")` & `post("/{project_id}:increase-priority")`

### DIFF
--- a/energetica/database/ongoing_project.py
+++ b/energetica/database/ongoing_project.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 from collections import defaultdict
 from dataclasses import dataclass, field
+from enum import IntEnum
 from functools import cached_property
 from typing import TYPE_CHECKING
 
@@ -11,12 +12,13 @@ from energetica import technology_effects
 from energetica.database import DBModel
 from energetica.enums import FunctionalFacilityType, ProjectType, TechnologyType, WorkerType
 from energetica.globals import engine
+from energetica.schemas.projects import ProjectOut
 
 if TYPE_CHECKING:
     from energetica.database.player import Player
 
 
-class ProjectStatus:
+class ProjectStatus(IntEnum):
     """Class that stores the status of ongoing projects."""
 
     PAUSED = 0
@@ -34,7 +36,7 @@ class OngoingProject(DBModel):
     duration: float  # in game ticks
     project_power: float  # Power consumed by the project
     project_pollution: float  # Emissions produced by the project
-    status: int = ProjectStatus.PAUSED  # 0 for paused, 1 for waiting, 2 for ongoing. See ProjectStatus
+    status: ProjectStatus = ProjectStatus.PAUSED  # 0 for paused, 1 for waiting, 2 for ongoing. See ProjectStatus
     end_tick_or_ticks_passed: float = 0  # in game ticks when the project will be finished or ticks passed if paused
 
     # multipliers to keep track of the technology level at the time of the start of the project
@@ -46,7 +48,8 @@ class OngoingProject(DBModel):
     def __post_init__(self) -> None:
         """Post initialization function."""
         self.multipliers = technology_effects.current_multipliers(
-            self.player, self.project_type
+            self.player,
+            self.project_type,
         )  # has to be called before the super().__post_init__()
         super().__post_init__()  # Needed for DBModel to add this new object to the database
 
@@ -59,6 +62,7 @@ class OngoingProject(DBModel):
         """Return True if this project is paused by the player."""
         return self.status == ProjectStatus.PAUSED
 
+    # TODO(mglst): make this method a property
     def is_ongoing(self) -> bool:
         """Return True if this project is not paused and has no requirements."""
         return self.status == ProjectStatus.ONGOING
@@ -191,3 +195,13 @@ class OngoingProject(DBModel):
                     if level + offset - 1 >= candidate_prerequisite_level:
                         prerequisites.append(candidate_prerequisite)
         return prerequisites, level
+
+    def to_schema(self) -> ProjectOut:
+        return ProjectOut(
+            id=self.id,
+            project_type=self.project_type,
+            end_tick=self.end_tick_or_ticks_passed if self.is_ongoing() else None,
+            ticks_passed=None if self.is_ongoing() else self.end_tick_or_ticks_passed,
+            duration=self.duration,
+            status=self.status,
+        )
--- a/energetica/database/ongoing_project.py
+++ b/energetica/database/ongoing_project.py
@@ -52,7 +52,7 @@ class OngoingProject(DBModel):
         """Return True if this project is paused by the player."""
         return self.status == ProjectStatus.PAUSED
 
-    # TODO(mglst): make this method a property
+    @property
     def is_ongoing(self) -> bool:
         """Return True if this project is not paused and has no requirements."""
         return self.status == ProjectStatus.ONGOING
@@ -60,7 +60,7 @@ class OngoingProject(DBModel):
     def pause(self) -> None:
         """Make this facility go from waiting or ongoing to paused."""
         assert not self.was_paused_by_player()
-        if self.is_ongoing():
+        if self.is_ongoing:
             self.end_tick_or_ticks_passed = self.duration - self.end_tick_or_ticks_passed + (engine.total_t + 1)
         self.status = ProjectStatus.PAUSED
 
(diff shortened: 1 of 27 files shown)


energetica/database/ongoing_project.py
```diff
@@ -59,6 +52,7 @@ def was_paused_by_player(self) -> bool:
         """Return True if this project is paused by the player."""
         return self.status == ProjectStatus.PAUSED
 
+    # TODO(mglst): make this method a property
```
mglst, line 55: TODO(mglst): make the `OngoingProject.is_ongoing` method into a property

energetica/database/player.py
```diff
@@ -551,6 +552,7 @@ def package_all() -> dict[int, dict]:
         """Package data for all players."""
         return {player.id: player.package() for player in Player.all()}
 
+    # TODO(mglst): deprecate this function
```
mglst, line 555: TODO(mglst): deprecate the `Player.package_constructions` method

energetica/routers/__init__.py
```diff
@@ -70,14 +73,23 @@ def validation_exception_handler(request: Request, exc: RequestValidationError)
     def global_exception_handler(request: Request, exc: GameError) -> JSONResponse:
         """Handle global game exceptions."""
         content = GameErrorOut.from_game_error(exc)
-        return JSONResponse(content=content.model_dump(), status_code=status.HTTP_403_FORBIDDEN)
+        return JSONResponse(content=content.model_dump(), status_code=status.HTTP_400_BAD_REQUEST)
```
mglst, line 76: The switch from 403 to 400 might cause issues elsewhere. Check this won't be an issue

energetica/routers/projects.py
```diff
@@ -0,0 +1,107 @@
...
+    player: Annotated[Player, Depends(get_current_user)],
+    project_id: int,
+    force: bool = False,
+) -> None:
+    """Cancel an ongoing project."""
+    project = OngoingProject.getitem(project_id, HTTPException(status_code=status.HTTP_404_NOT_FOUND))
+    if project.player != player:
+        raise HTTPException(status_code=status.HTTP_403_FORBIDDEN)
+    assets.cancel_project(player=player, project=project, force=force)
+    return None
```
mglst, line 55: This route should return `ProjectListOut`

energetica/schemas/common.py
```diff
@@ -2,9 +2,22 @@
...
+from pydantic import BaseModel, ConfigDict
+from pydantic.alias_generators import to_camel
+
+if TYPE_CHECKING:
+    from energetica.game_engine import Confirm
+    from energetica.game_error import GameError
+
+
+# TODO: switch to this as a base model
+class BaseApiModel(BaseModel):
```
mglst, line 16: Check that the new schemas inherit from `BaseApiModel`

energetica/schemas/common.py
```diff
@@ -15,3 +28,21 @@ class GameErrorOut(BaseModel):
     @classmethod
     def from_game_error(cls, game_error: GameError) -> GameErrorOut:
         return GameErrorOut(game_exception_type=game_error.exception_type)
+
+
+class ConfirmOut(BaseModel):
```
mglst, line 33: BaseApiModel

energetica/schemas/players.py
```diff
@@ -20,3 +20,9 @@ class UIStatePatch(BaseModel):
     """Model for updating the state of the UI."""
 
     last_opened_chat_id: int | None = Field(None, description="ID of the last opened chat")
+
+
+class MoneyOut(BaseModel):
```
mglst, line 25: `BaseApiModel`

energetica/static/frontend_data.js
```diff
@@ -64,11 +81,12 @@ function load_constructions() {
 }
 
 function retrieve_constructions() {
-    return fetch("/api/get_constructions")
+    return fetch("/api/v1/projects")
         .then((response) => response.json())
         .then((raw_data) => {
             // Save fetched data to sessionStorage
             sessionStorage.setItem("constructions", JSON.stringify(raw_data));
+            // TODO(mglst): rename the key used here. it should be projects not constructions
```
mglst, line 89: TODO(mglst): rename `"constructions"` to `"projectsData"`

energetica/utils/assets.py
```diff
@@ -460,14 +463,10 @@ def decrease_project_priority(player: Player, project: OngoingProject) -> None:
...
     """
-    if project is None or project.player != player:
-        msg = "Project not found"
-        raise GameError(msg)
-
     priority_list = player.projects_by_priority[project.project_type.worker_type]
     index = priority_list.index(project)
     if index == len(priority_list) - 1:
-        return
+        return  # TODO(mglst): raise a GameError
```
mglst, line 469: # TODO(mglst): raise a GameError

mglst (commented):
I did not review all of the JS

energetica/schemas/common.py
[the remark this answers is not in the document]
mglst: This requires some discussion - I've added a discussion point to the agenda of the next weekly meeting

energetica/utils/assets.py
```diff
@@ -496,12 +486,41 @@ def decrease_project_priority(player: Player, project: OngoingProject) -> None:
...
+    """
+    priority_list = player.projects_by_priority[project.project_type.worker_type]
+    index = priority_list.index(project)
+    if index == 0:
+        raise GameError("CannotIncreasePriorityOfFirstProject")
+    # Here, to increase the priority of this project, we decrease the priority of the project just above it
+    decrease_project_priority(player, priority_list[index - 1])
+
+
+def pause_project(player: Player, project: OngoingProject) -> None:
```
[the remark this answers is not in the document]
mglst: So regarding the 'necessity' of these functions, the question bubbles up to: is it necessary to have two separate routes? What was wrong with a unified toggle route? And my thinking here is that, well, they are separate actions, so they should be separate routes, purely out of principle. But more pragmatically, there can be other issues. For example, if a player has a bad network connection, and they pause a project, but nothing happens because of latency, so they after a few seconds of nothing happening, they press the button again. As a result, the project (eventually, once the server and network delays are over) gets first paused, and then resumed - because the same endpoint - toggle - was hit twice. With two separate endpoints, in this scenario, the endpoint pause would be called twice - the first would actually pause the project, the second would say "hey, there's an error, you asked to pause the project but it's already paused" which is better behavior. Really the best behavior would be to have the frontend show some kind of progress bar or spinner while it is waiting for the pause to actually take effect. But yeah those are my thoughts on the matter

energetica/utils/assets.py
```diff
@@ -496,12 +486,41 @@ def decrease_project_priority(player: Player, project: OngoingProject) -> None:
...
+    toggle_pause_project(player, project)
+
+
 def toggle_pause_project(player: Player, project: OngoingProject) -> None:
     """
     Pause or unpauses an ongoing project.
 
     Note : When a project is paused or unpaused, it's position in the priority list has to be updated.
     """
+    # TODO(mglst): this function should call pause_project and resume_project - that's where the logic should reside
```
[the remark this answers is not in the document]
mglst: If you look at the structure of the `toggle_pause_project` function, it's effectively a big case distinction: one where the project is paused, one where it's not. This effectively comes down to one case where we resume, and one case where we pause. And anyways - what is the use case for toggling? We only ever need to pause or resume, and we've been using toggle as a substitute for these.

energetica/static/progress_bar.js
```diff
@@ -289,45 +308,48 @@ load_constructions().then((constructions) => {
...
-  }, 100);
-});
+  });
+}, 100);
 
 function refresh_progressBar() {
-  load_constructions().then((construction_list) => {
+  // TODO(mglst): this function can REFRESH progress bars, but is not capable of CREATING or REMOVING html elements.
+  // e.g. this works when pausing and resuming projects, but not when queueing a new project or canceling one.
+  load_constructions().then((projectsData) => {
```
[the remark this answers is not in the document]
mglst: I've added a refresh to fix this - this was already the behavior before

energetica/utils/assets.py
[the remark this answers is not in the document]
mglst: Discussed with Felix, we agree this should be done. I will resolve this conversation when it is.

mglst:
@felixvonsamson in my view, this branch is ready to be merged.

PR #319, opened 2025-08-08
fix: add `method` to action logger and simulate

### LINKED ISSUE #314: Action logger should log HTTP method
This is a requirement now that not all methods are posts

mglst:
fixes #314


mglst:
This merge is superseded by the following merge: #321 
See the relevant issue #320

PR #325, opened 2025-08-11
refactor: added string enum `GameExceptionType`

mglst:
This pull request defines a string enum for all the game exception type. No actual modifications have been made here, rather, the exception types have simply been aggregated. This will now allow for simplification, now that they all appear together - we can make changes holistically.

### DIFF
--- a/energetica/routers/auth.py
+++ b/energetica/routers/auth.py
@@ -10,7 +10,7 @@ from fastapi import APIRouter, Depends, HTTPException, Request, Response, status
 from fastapi.responses import JSONResponse, RedirectResponse
 
 from energetica.database.player import Player
-from energetica.game_error import GameError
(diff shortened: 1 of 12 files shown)


energetica/routers/auth.py
```diff
@@ -62,7 +62,7 @@ def login(request_data: LoginRequest) -> Response:
 def signup(request: Request, request_data: SignupRequest) -> Response:
     """Create a new account."""
     if engine.disable_signups:
-        raise GameError("Sign-ups are disabled.")
+        raise GameError(GameExceptionType.TILE_NOT_FOUND)
```
felixvonsamson, line 65: I don't think that is the right exception

mglst: Good catch, thanks!

Issue #302, opened 2025-07-30
Issue with restarting server from instance


mglst:
<img width="1060" height="375" alt="Image" src="https://github.com/user-attachments/assets/6e4bc42b-f090-4843-8cde-cf35d1cf399c" />

mglst:
Steps to reproduce:
1. Start from a fresh instance
2. Create a new user by signing up and then immediately...
3. Stop the server with ctrl+C *before a tick occurs*
4. Assertion error `energetica/simulate.py` in `def create_user` line 24

Note if a tick occurs before stopping the server, the error does not occur.

The reason this bug exists is because the logic to determine if actions from the action history should be computed is based off the total_t. And since a tick has not taken place, the last actions after the last tick in the action history - here the `create_user` action - is run, but the player already exists, so when we try to simulate that action again, we get:
`raise HTTPException(status.HTTP_409_CONFLICT, "username is taken")`

@yassir-akram do you want to handle this?

mglst:
The crux of the issue is that using just `total_t` from the instance is not sufficient to determine which actions from the action history have been run and save to instance and which ones need to be simulated

mglst:
@yassir-akram any updates?

yassir-akram:
https://github.com/felixvonsamson/Energetica/commit/dddb39912a3dd2431cf3c931463ee663844491cf#diff-3e761df07d258284ee6f7cbd84de6ab25af21a65cd0578bcf54ba25f76d1f763R211 l211 is the problem...

mglst:
I get where you’re coming from — the current saving/checkpointing setup is definitely intricate and easy to break with small changes. That’s part of why I opened this MR in the first place.

From my side, I’ve tested this change and it does resolve the issue described here. I also think it moves us (in a small way) toward a cleaner structure that’s less brittle long term.

I’m not aiming to redesign the whole saving system right now — that’s a much bigger piece of work — but I do think this is a safe, targeted fix for the current bug, and I’d prefer not to just revert without considering whether this solves it more cleanly.

yassir-akram:
it resolves this particular problem and creates other ones! cf discussion we had a few months ago. Am happy to discuss this some other time, but it's time-consuming and brain-consuming. And not the priority for now. So for now I suggest we just move forward with the system that works. We can have a deeper conversation at some point.

mglst:
Got it — I don’t recall all the details from that earlier conversation, but fair enough on keeping priorities in check.

[...]

PR #350, opened 2025-08-13
feat: new `save_game_engine_instance` action to resume from

mglst:
This PR is my suggested solution to #302 

This adds a new action type, namely `save_game_engine_instance`, which is logged whenever the game engine's instance is saved. This can then be used in the logic to decide what actions, if any, need to be re simulated when resuming the server.

The logic for `actions_to_simulate` is not one I'm very familiar with. @yassir-akram could you give me your thoughts and feedback on my suggested addition and changes ?


mglst:
See #302 

Closing until revisited

PR #376, opened 2025-08-18
Docs/311-CONTRIBUTING.MD

### LINKED ISSUE #311: Create a `CONTRIBUTING.md`
It would also be good to write up an overview of the custom tools we've written for this project, like `simulate.py` and `init_test_players.py`

mglst:
Fixes #311


CONTRIBUTING.md
Entouanes, line 1: Excellent content. Here are a few suggestions:
- Adding a table of contents for navigation (higher priority)
- Break down the repo structure and explain each part (could be supported by an architecture diagram, e.g. mermaid chart) (medium priority)
- Adding diagrams for contribution workflow (e.g. mermaid chart) for a friendly visual aid (lower priority)

README.md
Entouanes, line 1: I would say "Running a local instance" should be kept in README.md for user who just intend to try out the project in their environment

felixvonsamson (approved):
I didn't real everything in detail, but it looks very good ! 
It might be a bit long in my opinion, some things could probably be left out.
I think we should merge this so that the documentation exists and we can review it in detail and update it if necessary in the future.

mglst:
I have the intention of looking into adding mermaid charts, as per Antoine's suggestion. I suppose the PR doesn't strictly need to wait for this.

PR #381, opened 2025-08-20
fix: add game verison to actions history and check version when running the log files

felixvonsamson:
I'm not sure I did this right, please check.

We probably want to have forward-compatible actions history. For now we can only run actions history of the same version (including beta and release candidates). What do you suggest ?

### DIFF
--- a/energetica/__init__.py
+++ b/energetica/__init__.py
@@ -1,7 +1,7 @@
 """Initializes the app and the game engine."""
 
-__version__ = "0.11.1-b"
-__release_date__ = "03/02/2025"
+__version__ = "0.11.2-b"
+__release_date__ = "20/08/2025"
 
 import asyncio
 import glob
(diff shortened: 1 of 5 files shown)


felixvonsamson:
Note that this makes previously generated actions_history.log files incompatible. This is why some tests fail.

energetica/__init__.py
```diff
@@ -133,11 +133,12 @@ def create_app(
         if actions:
             kwargs: dict = actions[0].model_dump()
             kwargs.pop("action_type")
-            # datetime.fromisoformat
-            kwargs["start_date"] = kwargs["start_date"]
```
mglst, line 137: This line was removed:
```
kwargs["start_date"] = kwargs["start_date"]
```
Is this an issue? @yassir-akram you'll know best

mglst (approved):

PR #437, opened 2025-08-28
fix: adjust lower and upper bounds for graphs

mglst:
Fix the min and max bounds for graph y auto scaling


mglst:
This PR is made redundant by #440

PR #438, opened 2025-08-28
fix: bugged progress bars

felixvonsamson:
This fixes the graphical bug with the progress bars when the server is restarted before the first checkpoint.


mglst (approved):
I am certainly no expert on the complex saving and checkpoints system, but from what I understand an from what I discussed with Felix, this seems alright by me

Issue #341, opened 2025-08-12
Network graph shows rounding error graph


felixvonsamson:
<img width="1175" height="397" alt="Image" src="https://github.com/user-attachments/assets/32b00525-f148-48f9-b2d7-d81551ea0877" />

Just after creating a network, I can see this graph for the import/import graph. This is due to rounding errors and is not critical but :
- The Network has just been created so there should not be a visible history. [TODO: discuss what the import export graph should show after switching to another network]
- It should be possible to get rid of the rounding errors.
- All graphs should have a minimal Y scale in order to not amplify rounding errors if there are any.

PR #440, opened 2025-08-30
fix: set very small values to 0 for net network exchanges

felixvonsamson:
This should fix the issue #341. Not convinced this is the best idea but it works.

### DIFF
--- a/energetica/production_update.py
+++ b/energetica/production_update.py
@@ -194,16 +194,22 @@ def calculate_net_import(new_values: dict) -> None:
     """
     exp = new_values["demand"]["exports"]
     imp = new_values["generation"]["imports"]
-    new_values["demand"]["exports"] = max(0.0, exp - imp)
-    new_values["generation"]["imports"] = max(0.0, imp - exp)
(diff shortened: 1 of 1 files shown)


energetica/production_update.py
```diff
@@ -194,16 +194,22 @@ def calculate_net_import(new_values: dict) -> None:
...
+    if abs(net_exchange) < 1:
+        net_exchange = 0.0
+    new_values["demand"]["exports"] = max(0.0, net_exchange)
+    new_values["generation"]["imports"] = max(0.0, -net_exchange)
     exp_rev = new_values["revenues"]["exports"]
     imp_rev = new_values["revenues"]["imports"]
+    net_exchange_revenues = exp_rev + imp_rev
+    if net_exchange == 0:
+        net_exchange_revenues = 0.0
     if exp_rev < 0 or imp_rev > 0:
```
mglst, line 207: Could you document this a bit? This case distinction. A comment for each branch maybe. Something that would also explain why we're taking min/min

Issue #435, opened 2025-08-26
Random Starting season for slightly better replayability


felixvonsamson:
Let's have the season at the start of the game not always be January so that the game is slightly more interesting to replay

PR #445, opened 2025-09-01
feat: use a random tick value for the delta_t to start at a random se…

felixvonsamson:
This adresses the feature request #435. delta_t is now asigned to a random value in order to start at a random moment in the in-game year.

### DIFF
--- a/energetica/game_engine.py
+++ b/energetica/game_engine.py
@@ -119,10 +119,9 @@ class GameEngine(object):
             disable_signups=disable_signups,
         )
         self.log_action(log_entry)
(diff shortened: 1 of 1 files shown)


mglst (changes requested):
For the other random but deterministic aspects of the game, we have used `np.random` rather than python's built in random library. I'm not clear on if one is better or more suitable than the other for our application, but certainly we should stick to just relying on one of them if we can, which in this situation we can, as far as I can see. As such, I would like to request that we make changes to use `np.random` to match the deterministic randomness as used for network prices, climate events, and 'location' of facilities in the tile for wind and solar facilities. These places can be found by searching for `"np.random"` as in string in the code base

PR #448, opened 2025-09-02
feat: changing urgent game values to balance the game

felixvonsamson:
This implements the urgent game value changes to try to balance the game a bit as discussed and agreed in the last meeting. 
I added also some rebalancing to the PV solar since the energy consumed for construction was quite extreme


yassir-akram (approved):
Super nice. Would be nice to have no CO2 absorbtion if its less than 300ppm. but not urgent.

mglst (approved):

Entouanes (approved):

felixvonsamson:
> Super nice. Would be nice to have no CO2 absorbtion if its less than 300ppm. but not urgent.

How do you justify this physically ?

Issue #312, opened 2025-08-04
Frontend graph problem


felixvonsamson:
<img width="1078" height="787" alt="Image" src="https://github.com/user-attachments/assets/dc8ad164-a264-45d6-837e-54a1fb34d92b" />

This was observed whith an industry and a construction when the industry has a higher energy consumpltion priority than the constrcuctions and specifically for the moments when the power demand of the construction is reduced due to energy shortage

felixvonsamson:
<img width="1078" height="787" alt="Image" src="https://github.com/user-attachments/assets/bca0313a-0559-4de0-81ef-69345c1ff50f" />

Another situation where the construciton where slowed down. I am looking at the electricity graph

Issue #370, opened 2025-08-18
Paused project shouldn't have a snail icon


yassir-akram:

mglst:
This is likely related to #334

Issue #334, opened 2025-08-12
Construction that are stopped due to lack of electricity still progress


felixvonsamson:
<img width="1183" height="279" alt="Image" src="https://github.com/user-attachments/assets/586afa28-5b5b-42ab-a01b-ad8c9acd3b4f" />

The two constructions on the screenshot should not be able to progress because of 0% power satisfaction. They still progress in chunks somehow. Probably only a frontend problem since a refresh fixes the progress status

mglst:
This is likely related to #370

PR #449, opened 2025-09-04
Fix/312+334+370

felixvonsamson:
these bug fixes should fix the issues #312, #334 and #370.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -11,9 +11,9 @@
             "program": "main.py",
             "console": "integratedTerminal",
             "args": [
-                "--env", "prod", //maybe add a new test environment
+                "--env", "dev", //maybe add a new test environment
                 "--port", "5001",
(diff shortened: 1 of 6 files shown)


.vscode/launch.json
```diff
@@ -11,9 +11,9 @@
             "program": "main.py",
             "console": "integratedTerminal",
             "args": [
-                "--env", "prod", //maybe add a new test environment
```
mglst, line 14: Is there a reason for the changes to the `launch.json` file? I would rather these stay as they are

mglst (approved):
Great stuff. Did you test that the three issues are actually resolved?

felixvonsamson:
> Great stuff. Did you test that the three issues are actually resolved?

Yes I did adress all 3 issues. Didn't check all possible scenarios since we could think of a lot but i think the pre-release can be used for this.

PR #452, opened 2025-09-05
fix: displaying the correct upgrade values for technologies

felixvonsamson:
this PR fixes the display of the upgrade data seen on the technology page. That has been forgotten after the implementation of the special_multiplier to make the progression less exponential.

Note: I am not happy with the way the upgrade system works but this should be changes in the context of a broader discussion about the games values.

### DIFF
--- a/energetica/technology_effects.py
+++ b/energetica/technology_effects.py
@@ -43,6 +43,21 @@ def special_multiplier(pf: float, lvl: int) -> float:
     return (0.5 + 0.5 * pf) ** lvl + np.log(pf / (0.5 + 0.5 * pf)) * lvl
 
 
+def special_multiplier_gain(pf: float, lvl: int) -> float:
+    """Returns the value multiplier for one additional level in percent."""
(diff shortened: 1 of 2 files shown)


energetica/technology_effects.py
```diff
@@ -318,11 +332,12 @@ def efficiency_multiplier_for_storage_facilities(
 
 def extraction_emissions_multiplier(player: Player, extraction_facility_type: ExtractionFacilityType) -> float:
     """Return by how much the `facility`'s `base_pollution` should be multiplied."""
+    # TODO(Felix): The if is is redundant no, this function is always called for extraction facilities.
```
mglst, line 335: Regarding the TODO: I would argue that if the `if` is redundant, then that makes the `affected_facilities` key for `mineral_extraction` redundant too

mglst (approved):

energetica/technology_effects.py
felixvonsamson: well in that case yes

felixvonsamson: I dont know, but I would say no for consistency

PR #469, opened 2025-09-12
fix: update logic for actions regarding users and players

### LINKED ISSUE #468: Player_id for account creation does not match with player_id for the logged actions
Here you can see an extract from an actions_history.log with the creation of an account with player_id=2 and the choice of location from the same player with a player_id=1. 
```
{"timestamp":"2025-09-06T11:33:43.337203","ip":"217.22.139.31","action_type":"create_user","player_id":2,"username":"Felix","pw_hash":"$2b$12$yz.Sq7rVzAeqz42Lt95kEOD6QiGOQ5uOBkjaqFFlpixwgtiklMVaS"}
{"timestamp":"2025-09-06T11:34:00.001473","action_type":"tick","total_t":2,"elapsed":0.000456}
{"timestamp":"2025-09-06T11:34:23.380680","elapsed":0.003172,"ip":"217.22.139.31","action_type":"request","player_id":1,"request":{"endpoint":"/api/v1/map/226:settle","method":"POST","content_type":"application/json","payload":{}},"response":{"status_code":204,"content_type":"application/json","payload":"unparsable or not JSON"}}
```

It is worth noting that this does not produce any unexpected behaviour for a normal run of the game (non simulated)

mglst:
* when a user settles on a tile, it creates a player. As such, this is no longer considered an API action like the rest, but acts like an auth action. As such:
* there is now a new CreatePlayerAction (which is distinct from the CreateUserAction)
* as was already the case, all API actions must be originated by an authenticated user, who has an assocaited player. The player id for these actions only are stored. Therefore it is normal for the first player's actions to be recorded with an id of 1, since that is where id's start counting from, even though their user is is 2, or anything else

> [!WARNING]
> For migrating an action log that was generated before this change, settle actions need to have an additional `player_id` response

Resolves #468 #462 #461

### DIFF
--- a/energetica/routers/auth.py
+++ b/energetica/routers/auth.py
@@ -15,7 +15,6 @@ from energetica import __release_date__, __version__
 from energetica.database.user import User
 from energetica.game_error import GameError, GameExceptionType
 from energetica.globals import engine
-from energetica.routers.map import router
 from energetica.schemas.auth import (
     ChangePasswordRequest,
     LoginRequest,


energetica/simulate.py
```diff
@@ -96,8 +98,21 @@ def _simulate(
...
             user_sessions[player_id] = create_user(player_id, action.username, action.pw_hash)
+        elif action.action_type == "create_player":
+            user = User.getitem(
+                action.user_id,
+                ValueError(f"Cannot create player: user with id {action.user_id} does not exist"),
+            )
+            if user.player is not None:
+                print("\033[31mUser already has a player.\033[0m")
+                break
+            tile = HexTile.getitem(
```
felixvonsamson, line 111: I think this should be a tile_id

energetica/schemas/simulate.py
```diff
@@ -24,11 +24,18 @@ class CreateUserAction(BaseModel):
     timestamp: datetime
     ip: str | None
     action_type: Literal["create_user"]
-    player_id: int
+    user_id: int
     username: str
     pw_hash: str
 
 
+class CreatePlayerAction(BaseModel):
```
felixvonsamson, line 32: This action should include the tile_id

energetica/schemas/simulate.py
```diff
@@ -24,11 +24,18 @@ class CreateUserAction(BaseModel):
     timestamp: datetime
     ip: str | None
     action_type: Literal["create_user"]
-    player_id: int
+    user_id: int
```
felixvonsamson, line 27: will this change not have repercussion for the simulation ?

felixvonsamson (changes requested):

mglst:
> I didn't see anything about the removal of the logged action for settle

There's an added condition in the `routers/__init__.py', added to the check for auth routes.

energetica/schemas/simulate.py
mglst: Yes - for previous versions of action history. I believe my warning in the PR description should address this

mglst:
Good catch @felixvonsamson - I'll resolve

felixvonsamson (approved):

PR #478, opened 2025-09-27
Refactor/const-config-bis

### LINKED ISSUE #447: Missing `danger_description` for coal burner

### LINKED ISSUE #472: Poor documentation for efficiency / possible duplicate field for storage facilities in config
Image

See image above. Storage facilities have both a `base_efficiency` and a `initial_efficiency` field.

I recall having a conversation about there really being two efficiencies:
1. the efficiency of converting electric energy into stored energy
2. the efficiency of converting stored energy back into electric energy
This distinction is indeed relevant, since it affects all computations, and it is realistic to model things this way

However, I also recall a conversation along the lines of: "it's fine, we just need to model the overall efficiency, so take the square root of the desired efficiency on each end". 

I do not know which of these two scenarios the code currently is in. Either way:
* if there is a duplicate, one of them should be removed - which one?
* if it is not a duplicate, these should be renamed. To what?

@felixvonsamson

mglst:
resolves #472
resolve #447

### DIFF
--- a/energetica/config/assets.py
+++ b/energetica/config/assets.py
@@ -281,7 +281,6 @@ const_config: dict = {
             "ramping_time": timedelta(minutes=9).total_seconds(),  # [in-game seconds]
             "lifespan": timedelta(days=525).total_seconds(),  # [in-game seconds]
             "image_extension": "png",
-            "initial_efficiency": 0.75,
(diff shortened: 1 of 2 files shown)


felixvonsamson:
you renamed initial_efficiency to base_efficiency but I dont see any changes in the config. Was it already changed previously ?

PR #489, opened 2025-10-15
fix: fetch relevant data to prevent choosing a location clash (#486)

### LINKED ISSUE #486: Map is not dynamically updating when players are choosing a location and map display bugs
image

mglst:
fix #486


felixvonsamson:
Yes that looks like a good solution but it would have made it even worse because of the (still undefined) bug for the map display that would chrash at the first user and only display a few tiles. This is probably due to a badly managed session storage and is an issue that is present since a very long time.

mglst:
Good to merge?

PR #583, opened 2026-03-30
fix: restore logout and settle redirect broken by #568

### LINKED ISSUE #576: Logging out does not work properly
Steps to rerpoduce: log out.

I suspect this is related to #568

### LINKED ISSUE #579: After settleing, the player is not redirected to the dashboard

felixvonsamson:
## Summary

PR #568 migrated `AuthContext` from `useState` to `useQuery`, which inadvertently broke two flows.

- **fix #576** — `logout()` in `auth-context.tsx` was changed from `setUser(null)` to `queryClient.clear()`. `clear()` removes the auth cache entry and triggers a background refetch, which could briefly restore the session if the cookie wasn't cleared yet, and caused `isLoading = true` to flash (unmounting `LogoutComponent` mid-flow). Fixed by using `queryClient.setQueryData(queryKeys.auth.me, null)` — the direct equivalent of the old `setUser(null)`.

- **fix #579** — `handleSettleLocation` called `await refetchAuth()` then `navigate({ to: "/app/dashboard" })` immediately. `refetch()` resolves when the HTTP request completes and the cache is updated, but React hasn't re-rendered `AuthProvider` yet. The root guard saw stale `user.is_settled = false` on the dashboard route and redirected back to `/app/settle`. Fixed by removing the explicit `navigate()` and replacing it with a `useEffect` in `SettleContent` that watches `user?.is_settled` — it only fires after React has re-rendered with the new auth state.

## Test plan

- [ ] Log out → should land on `/app/login` with no flash or redirect loop
- [ ] Settle a location → should be redirected to `/app/dashboard` immediately after settling

<h3>Greptile Summary</h3>

This PR fixes two regressions introduced by #568's migration from `useState` to `useQuery` for auth state: the logout flow (which flashed a loading state due to `queryClient.clear()` triggering an immediate refetch) and the settle-location redirect (which raced between the HTTP response completing and React re-rendering with the updated auth state).

Key changes:
- **Logout**: `queryClient.clear()` replaced with `queryClient.setQueryData(auth.me, null)` + `queryClient.removeQueries(...)` in `useLogout`, and the `LogoutComponent` is rewritten to use the mutation hook with a `useRef` guard against React StrictMode double-invocation.
- **Settle**: `TileInfo` now calls `queryClient.invalidateQueries(auth.me)` after a successful settlement instead of `await refetchAuth() + navigate(...)`. `SettlePage` wraps `SettleContent` and reactively navigates to `/app/dashboard` once `user.is_settled` becomes `true`, ensuring the navigation happens after React re-renders with the new auth state.
- **Root guard**: Extracted redirect logic into a pure `computeRedirect` function, reducing the effect to a single `navigate` call, and added a render-block (`return null`) while the redirect fires.
- **Backend middleware**: `/auth/me` is excluded from the server-side 401→`/app/login` redirect, so the frontend `useQuery` receives a proper 401 and can return `null` instead of following a redirect.

Safe to merge after fixing the unhandled promise rejection in the logout path; the settle fix and root guard refactor are solid.

One P1 remains: `logout.mutateAsync()` without a `.catch()` produces an unhandled promise rejection on API failure, which can surface as a console error or trigger `unhandledrejection` listeners. All other findings are P2 (debug-style loading strings). The two core fixes (logout flash and settle race) are correctly implemented.

frontend/src/routes/app/logout.tsx — unhandled rejection from mutateAsync

| Filename | Overview |
|----------|----------|
| frontend/src/routes/app/logout.tsx | Refactored to use `useLogout` mutation hook; adds StrictMode guard via `useRef`, but `mutateAsync` rejection is unhandled — `.finally()` navigates to login regardless, yet the unhandled promise surfaces as an `UnhandledPromiseRejection` on API failure. |
| frontend/src/hooks/use-auth-queries.ts | Logout mutation now correctly uses `setQueryData(null)` instead of `clear()` to avoid the refetch race, and removes other user-specific cache entries with `removeQueries`; `mutationFn` now properly awaits the API call. |
| frontend/src/contexts/auth-context.tsx | Cleanly removes the `logout` method and `queryClient` import; logout responsibility is now fully in `useLogout` mutation hook. |
| frontend/src/routes/__root.tsx | Root guard refactored into a pure `computeRedirect` function with a minimal `useEffect`; debug-style strings (`"isLoading (useAuth())"`, `"capabilities === undefined"`) are user-visible loading placeholders. |
| frontend/src/routes/app/settle.tsx | Settlement flow fixed: `TileInfo` now invalidates the auth query instead of awaiting `refetchAuth()` + imperative navigate; `SettlePage` watches `user?.is_settled` reactively and navigates to dashboard once settled. |
| energetica/routers/__init__.py | Backend middleware no longer redirects `/auth/me` 401s to `/app/login`, allowing the frontend `useQuery` to receive the expected 401 and return `null` cleanly. |

</details>

<sub>Reviews (3): Last reviewed commit: ["fix: redirect on settle"](https://github.com/felixvonsamson/energetica/commit/1c135ea1c790678749a4d95b1b94a9ba538be65b) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=26830159)</sub>

### DIFF
--- a/frontend/src/routes/app/logout.tsx
+++ b/frontend/src/routes/app/logout.tsx
@@ -16,10 +16,13 @@ function LogoutComponent() {
         if (hasLoggedOut.current) return;
         hasLoggedOut.current = true;
 
-        logout.mutateAsync().finally(() => {
-            navigate({ to: "/app/login" });
-        });
-    // eslint-disable-next-line react-hooks/exhaustive-deps
+        logout
+            .mutateAsync()
+            .catch(() => {})
+            .finally(() => {
+                navigate({ to: "/app/login" });
+            });
+        // eslint-disable-next-line react-hooks/exhaustive-deps
     }, []);
 
     return <div>Logging out...</div>;


mglst:
The long loading times is because the query fails, and TanStack query tries 5 times with a exponential backoff, before eventually giving up, which is why we're seeing the long loading times

mglst:
Since we're using cookie based auth, logging out should clear this cookie (in `frontend/src/routes/app/logout.tsx` or `frontend/src/contexts/auth-context.tsx`) but right now this does not happen, so logout effectively fails

mglst:
Okay - I think the issue is that unathenticated requests to `api/v1/auth/me` are incorrectly being responded to with `303` reesponses instead `401` responses.

PR #632, opened 2026-04-05
feat: add bankruptcy expulsion

mglst:
<h3>Greptile Summary</h3>

This PR replaces the old "credit limit exceeded" notification and demand-reduction logic with automatic network expulsion: players whose balance drops below `-income_per_day` are ejected from their network at the start of each tick, receive a `network_expelled` notification, and are blocked from rejoining until their balance is positive again.

- **Backend**: Expulsion loop added in `update_electricity()` before market resolution; the market loop now uses a fresh `Network.all()` call (fixing the previously-flagged stale-network iteration bug); demand-reduction code and the old `CreditLimitExceededPayload` schema are removed.
- **Frontend**: Create/join market dialogs show an "Insufficient funds" banner and disable their primary action when `player.money ≤ 0`; notification config and generated types updated for the new payload.
- **Tests**: New unit test guards against `NotificationType` and `PersistableNotificationPayload` falling out of sync in the future.
- **Docs**: New wiki section explains the join/create restriction and expulsion rule, though the stated threshold ("below zero") does not match the code's actual threshold (`-income_per_day`).

Safe to merge; expulsion logic is correct and both previously flagged P1 concerns are fully resolved

Both P1 concerns from the prior review (stale networks list and stale generated types) are fully addressed. The expulsion loop correctly uses list(network.members) to copy before iterating, captures network_name before calling leave_network, and uses a fresh Network.all() call for the subsequent market-resolution loop. The only remaining finding is a P2 wiki documentation inaccuracy where the described threshold (zero) differs from the code's actual threshold (-income_per_day). This does not affect runtime behaviour and does not block merge.

frontend/src/content/wiki/network.mdx — expulsion threshold description should be updated to match the -income_per_day code threshold

| Filename | Overview |
|----------|----------|
| energetica/production_update.py | Adds expulsion loop before market resolution; fresh Network.all() call prevents stale-iteration bug; demand-reduction code correctly removed |
| energetica/utils/network_helpers.py | Unchanged in this PR; leave_network correctly sets player.network=None and deletes empty networks |
| energetica/schemas/notifications.py | CreditLimitExceededPayload replaced by NetworkExpelledPayload with network_name field |
| energetica/database/messages.py | NotificationType literal updated from credit_limit_exceeded to network_expelled |
| frontend/src/lib/notification-config.tsx | Notification config updated with expulsion message and correct redirect to electricity-markets page |
| frontend/src/types/api.generated.ts | Generated types updated: CreditLimitExceededPayload removed, NetworkExpelledPayload with network_name added to union |
| frontend/src/components/electricity-markets/create-market-dialog.tsx | Adds insufficient-funds banner and disables create button when player balance is at or below zero |
| frontend/src/components/electricity-markets/join-market-dialog.tsx | Adds insufficient-funds banner and disables join button when player balance is at or below zero |
| frontend/src/content/wiki/network.mdx | New bankruptcy section added but states expulsion threshold as zero rather than the code's actual -income_per_day threshold |
| tests/unit/test_notification_types.py | New test ensures NotificationType and PersistableNotificationPayload stay in sync; prevents future regressions like the one this PR fixes |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant UE as update_electricity()
    participant N as Network.all()
    participant P as Player
    participant NH as leave_network()
    participant MR as Market Resolution

    UE->>N: list(Network.all()) — expulsion pass
    loop For each network
        loop For each member
            UE->>P: check money < -income_per_day
            alt Bankrupt
                UE->>NH: leave_network(player)
                NH-->>UE: network deleted if now empty
                UE->>P: notify(NetworkExpelledPayload)
            end
        end
    end
    UE->>N: Network.all() — fresh call for market resolution
    loop For each surviving network
        UE->>MR: calculate demand, generation, market logic
        MR-->>UE: market price and quantities
        UE->>P: emit new_network_values
    end
```

<sub>**[Greploops](https://www.greptile.com/docs/mcp-v2/skills#greploop)** — Automatically fix all review issues by running `/greploops` in [Claude Code](https://claude.ai/download). It iterates: fix, push, re-review, repeat until 5/5 confidence.<br>Use the **[Greptile plugin for Claude Code](https://www.greptile.com/docs/integrations/claude-code#claude-code-plugin)** to query reviews, search comments, and manage custom context directly from your terminal.</sub>

<sub>Reviews (2): Last reviewed commit: ["fix: simpler wording for wiki about bank..."](https://github.com/felixvonsamson/energetica/commit/e89893c0ef850752ee9dee90915321334cb8b828) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=27375328)</sub>

### DIFF
--- a/frontend/src/content/wiki/network.mdx
+++ b/frontend/src/content/wiki/network.mdx
@@ -49,6 +49,6 @@ This automatic process ensures fair transactions based on market supply and dema
 
 ## Bankruptcy and Network Expulsion
 
-Players can only join or create a network if they have a positive balance before.
+You can only join or create a network if your money balance is positive.
 
-If a player's money balance falls below zero, they are automatically removed from their network.
+If a player in a network falls below one day of industry income in debt, they are automatically removed from the network.


energetica/utils/network_helpers.py
```diff
@@ -15,6 +15,8 @@ def join_network(player: Player, network: Network | None) -> Network:
         # TODO(mglst): this logic should be handled by the router so that it can return a relevant HTTP error code such
         # as HTTP_404_NOT_FOUND
         raise GameError(GameExceptionType.NO_SUCH_NETWORK)
+    if player.money <= 0:
+        raise GameError(GameExceptionType.NOT_ENOUGH_MONEY)
```
felixvonsamson: This game exception tells the player "You don't have enough money for this action." Which is convenient because it works for everything but would we want to be more precise with this message. "You don't have enough money to start this Project" for projects and "You need to have a positive balance to join a Network" for this here.
Not a must, it works like this

energetica/utils/network_helpers.py
```diff
@@ -15,6 +15,8 @@ def join_network(player: Player, network: Network | None) -> Network:
         # TODO(mglst): this logic should be handled by the router so that it can return a relevant HTTP error code such
```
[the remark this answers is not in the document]
mglst: Not really. What I'm pointing at here is the semantic difference between HTTP error codes. When `network is None`, which means `NO_SUCH_NETWORK` then really this is like accessing a page that does not exist, and there is a dedicated HTTP code for that - `404`. 

Currently, we're doing `raise GameError(GameExceptionType.NO_SUCH_NETWORK)` - and all `GameError`s get turned into generic `400` errors. My point is that we should use `404` instead of `400` + a 'custom' error. And indeed, if we switched over to using just `404`, we could get rid of `NO_SUCH_NETWORK` entirely. And indeed, we could generally clean up `GameExceptionType` even further. We could get rid of:
- `TILE_NOT_FOUND`
- `PROJECT_NOT_FOUND`
- `FACILITY_NOT_FOUND`
- `USER_NOT_FOUND`

Effectively, this is what we're already doing in `energetica/routers/chats.py`, note also the `403`:
```py
@router.get("/{chat_id}/messages")
def get_chat_messages(
    player: Annotated[Player, Depends(get_settled_player)],
    chat_id: int,
) -> MessageListOut:
    """Get the messages for a chat."""
    chat = Chat.getitem(chat_id, error=HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail="Chat not found"))
    if player not in chat.participants:
        raise HTTPException(status_code=status.HTTP_403_FORBIDDEN, detail="You are not a participant in this chat")
    return MessageListOut(messages=[MessageOut.from_message(message) for message in chat.messages])

```
`energetica/routers/chats.py:28-38`

So yeah we _could_ do some further cleanup and homogenisation of error handling by making better use of the semantics of HTTP error codes

mglst:
@greptileai

frontend/src/content/wiki/network.mdx
```diff
@@ -46,3 +46,9 @@ The electricity market operates automatically and follows specific steps:
 4. **Transaction:** Capacities below the market price are sold at market price, while demands above the market price are bought at market price.
 
 This automatic process ensures fair transactions based on market supply and demand dynamics.
+
+## Bankruptcy and Network Expulsion
+
+Players can only join or create a network if they have a positive balance before.
+
+If a player's money balance falls below zero, they are automatically removed from their network.
```
felixvonsamson: Haha, this is indeed simpler wording but that is not true ^

felixvonsamson (approved):

Issue #752, opened 2026-05-05
Show monthly capacity factor sparkline in facility dialog for solar and hydro


mglst:
## Problem

Players are investing in solar (PV, CSP) and hydro facilities without understanding their seasonal output profile. Solar in particular is dramatically weaker in winter — players have reported panels never exceeding 30% capacity and CSP mirrors barely covering their own maintenance costs — but only discover this after building. The seasonal variation is not communicated at decision time.

## Proposed solution

Add a **12-bar sparkline** (one bar per in-game month) to the facility stats dialog for solar and hydro facilities, showing the expected capacity factor by month at the player's tile location.

- **Solar:** derived from the clear-sky irradiance model already used by the engine (astronomical calculation, fully deterministic given tile coordinates). No live data needed — compute once per tile and cache.
- **Hydro:** equivalent seasonal discharge/flow profile for the player's tile.

The sparkline sits inline with the other output stats (power, O&M cost, etc.) — compact enough not to clutter the dialog, but enough to immediately convey "this is a summer technology" or "this river runs dry in August."

## What this is not

Not a link to the weather widget (which shows current conditions, not seasonal expectations) and not a full interactive chart. Just a small, static, read-at-a-glance bar chart.

## Source

Raised by players in general chat after expressing confusion about poor solar performance in winter (May 5, 2026).

felixvonsamson:
I like the proposed solution. My only concern is that this could be seen as too much information. To solve this issue we could 'hide' this information behind a button but if you think otherwise I don't mind.

PR #757, opened 2026-05-06
Show installed capacity and capacity factor for storage in power overview

### LINKED ISSUE #643: installed capacity and usage not shown for storage facilities
The problem here is we have to decide if we want to show the installed capacity and usage of storage or power or maybe both. 
I think power makes more sense.

mglst:
## Summary
- Storage facilities now populate the "Installed Cap." and gauge columns on the generation tab of the power overview, using their `max_power_generation` (max discharge rate).
- Renamed "Used Capacity" to "Capacity Factor" since the gauge shows `avg_power / max_power` as a percentage — the textbook definition of capacity factor.

Note: storage's discharge-side capacity factor will look structurally low (a battery doing daily peak-shaving spends most of its time idle or charging), but the number is well-defined and consistent with the column on this tab. Extending the consumption tab with the symmetric charge-side metric is left as future work.

Closes #643

## Test plan
- [ ] Open `/app/overviews/power` with a game that has at least one storage facility (e.g. small pumped hydro storage).
- [ ] On the **generation** tab, confirm the storage row now shows a value in the "Installed Cap." column and a populated gauge in the "Capacity Factor" column.
- [ ] Confirm the column header reads "Capacity Factor" (not "Used Capacity") and that sorting on that column still works.
- [ ] Non-storage rows (gas, solar, etc.) are unchanged.
- [ ] Consumption tab is unchanged.

<h3>Greptile Summary</h3>

Storage facilities now populate the "Installed Cap." and "Capacity Factor" columns on the generation tab of the power overview by spreading `storage_facilities` into the same array used for `power_facilities` during the capacity lookup. The column header is also renamed from "Used Capacity" to "Capacity Factor" with a dotted-underline tooltip that defines the metric and links to the wiki.

- **Installed capacity & capacity factor for storage**: `storage_facilities` are spread into the combined array filtered by `facilityType`, using `max_power_generation` (discharge rate) for both installed-capacity aggregation and the capacity-factor calculation — consistent with how `power_facilities` are handled.
- **Column rename + tooltip**: "Used Capacity" → "Capacity Factor" with a Radix `Tooltip` explaining `avg_power / installed_capacity`; the `Link` inside `TooltipContent` renders in a portal so clicks do not accidentally trigger the sort handler on the `<th>`.

Safe to merge — the change is a small, well-contained addition that reuses the same field access pattern already established by power_facilities.

Both StorageFacilityOut and PowerFacilityOut expose facility and max_power_generation, so spreading them into the same array is type-safe and the capacity calculation is correct. The only open question is whether the Learn more wiki link targets the most helpful wiki page, but this does not affect correctness or runtime behavior.

| Filename | Overview |
|----------|----------|
| frontend/src/components/charts/power-chart.tsx | Adds storage facilities to the capacity/capacity-factor lookup in the generation table, and renames "Used Capacity" to "Capacity Factor" with an explanatory tooltip. Logic and types are correct; both `StorageFacilityOut` and `PowerFacilityOut` expose `facility` and `max_power_generation`. Minor: the tooltip's "Learn more" link targets the `power-facilities` wiki slug even for storage rows. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A[chartData keys → facilityTypes] --> B{isGeneration && facilitiesData?}
    B -- No --> C[row: facilityType + totalEnergy only]
    B -- Yes --> D["[...power_facilities, ...storage_facilities]\n.filter(f => f.facility === facilityType)"]
    D -- "facilities.length === 0" --> C
    D -- "facilities.length > 0" --> E["installedCapacity = Σ max_power_generation"]
    E --> F["avgPower = totalEnergy / (ticks × timeInHours)"]
    F --> G["usedCapacity = avgPower / installedCapacity × 100"]
    G --> H[FacilityGauge value=usedCapacity]
    C --> I[Show '-' in Installed Cap. & Capacity Factor columns]
```

<sub>Reviews (3): Last reviewed commit: ["fix: full-text tooltip with wiki link on..."](https://github.com/felixvonsamson/energetica/commit/765d502ac62987a2d8786b9e5d6a73596b8fa141) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=31054278)</sub>

### DIFF
--- a/frontend/src/components/charts/power-chart.tsx
+++ b/frontend/src/components/charts/power-chart.tsx
@@ -8,6 +8,7 @@
  *  - Facility icon + name labels in the tooltip
  */
 
+import { Link } from "@tanstack/react-router";
 import { useMemo, useState } from "react";
 
 import {
@@ -329,10 +330,19 @@ export function PowerOverviewTable({
                                                 Capacity Factor
                                             </span>
                                         </TooltipTrigger>
-                                        <TooltipContent>
-                                            Avg. power output over the
-                                            displayed time range ÷ installed
-                                            capacity
+                                        <TooltipContent className="max-w-52 text-center leading-snug">
+                                            Average power output over the
+                                            displayed time range divided by
+                                            installed capacity.{" "}
+                                            <Link
+                                                to="/app/wiki/$slug"
+                                                params={{
+                                                    slug: "power-facilities",
+                                                }}
+                                                className="underline hover:opacity-80"
+                                            >
+                                                Learn more
+                                            </Link>
                                         </TooltipContent>
                                     </Tooltip>
                                     {getSortIndicator("used")}
--- a/frontend/src/components/charts/power-chart.tsx
+++ b/frontend/src/components/charts/power-chart.tsx
@@ -17,6 +17,11 @@ import {
 import { FacilityIcon } from "@/components/ui/asset-icon";
 import { FacilityName } from "@/components/ui/asset-name";
 import { FacilityGauge } from "@/components/ui/facility-gauge";
+import {
+    Tooltip,
+    TooltipContent,
+    TooltipTrigger,
+} from "@/components/ui/tooltip";
 import { useAssetColorGetter } from "@/hooks/use-asset-color-getter";
 import { useFacilities } from "@/hooks/use-facilities";
 import { useGameEngine } from "@/hooks/use-game";
@@ -318,7 +323,18 @@ export function PowerOverviewTable({
                                     className="py-3 px-4 text-center font-semibold cursor-pointer hover:bg-tan-green/80 dark:hover:bg-card transition-colors min-w-37.5"
                                     onClick={() => handleSort("used")}
                                 >
-                                    Capacity Factor
+                                    <Tooltip>
+                                        <TooltipTrigger asChild>
+                                            <span className="underline decoration-dotted cursor-help">
+                                                Capacity Factor
+                                            </span>
+                                        </TooltipTrigger>
+                                        <TooltipContent>
+                                            Avg. power output over the
+                                            displayed time range ÷ installed
+                                            capacity
+                                        </TooltipContent>
+                                    </Tooltip>
                                     {getSortIndicator("used")}
                                 </th>
                             </>
--- a/frontend/src/components/charts/power-chart.tsx
+++ b/frontend/src/components/charts/power-chart.tsx
@@ -213,9 +213,10 @@ export function PowerOverviewTable({
                 const row: FacilityRow = { facilityType, totalEnergy };
 
                 if (isGeneration && facilitiesData) {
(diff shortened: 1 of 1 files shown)


frontend/src/components/charts/power-chart.tsx
```diff
@@ -317,7 +318,7 @@ export function PowerOverviewTable({
                                     className="py-3 px-4 text-center font-semibold cursor-pointer hover:bg-tan-green/80 dark:hover:bg-card transition-colors min-w-37.5"
                                     onClick={() => handleSort("used")}
                                 >
-                                    Used Capacity
+                                    Capacity Factor
```
felixvonsamson: Capacity factor means something else and I would avoid using this wording here since it might be confusing.

The capacity factor is usually referring to the average production of a power plant over a long period divided by the maximal theoretical porduction of this power plant over that period. It is usually used for renewable power plants to show the difference between installed capacity and real power output. 

example for wind turbine:

Image

felixvonsamson (changes requested):
Revert renaming of "Used capacity" or find another wording

frontend/src/components/charts/power-chart.tsx
mglst: but isn't this precisely what the gauge is showing?
> average production of a power plant over a long period divided by the maximal theoretical porduction of this power plant over that period

felixvonsamson: The gauge is showing how much the facility is currently producing compared to the installed capacity. This gives no information about how much electricity you could get from this facility over a long period knowing the amount of installed capacity as opposed to the capacity factor in the facility dialog.

The capacity factor is the average value of the "current usage" of a power plant over a long period (and is more relevant for renewables). 

The graph suggested in #752 will represent capacity factors but the "current usage" of a solar facility will be 0% during the night and will vary a lot.

mglst: I don't understand. I don't see an issue? I think this is precisely how I had it in mind. This is underneath the graph for her power facilities so it shows power levels.

felixvonsamson: The Gauge in this column is not showing the instantaneous capacity usage like the column "output" in the management page but the produced power in the displayed range over `installed_capacity * time_range`. This is indeed a kind of capacity factor but on a variable period. 
Ok to write "Capacity factor" but probably with an asterix and a hover info "over the displayed time range" and a link to a more detailed explanation in the wiki.

mglst: Done — added a dotted-underline tooltip to the header with the text "Avg. power output over the displayed time range ÷ installed capacity".

mglst:
image
See also #778 as a follow up, for a more precise wiki entry abotu capacity factor.

mglst:
See also #768

felixvonsamson (approved):

PR #765, opened 2026-05-07
fix: prevent dict-changed-during-iteration crash in DBModel.all()

felixvonsamson:
## What happened

Today on `energetica-edu`, the server entered a crash loop caused by two separate issues that compounded each other.

### 1. Race condition: `RuntimeError: dictionary changed size during iteration`

Seen twice in production logs (12:41 and 15:17):

```
File "energetica/database/__init__.py", line 86, in <genexpr>
    return (v for v in cls.instances().values())
RuntimeError: dictionary changed size during iteration
```

**Root cause:** `DBModel.all()` returned a lazy generator over the live `instances()` dict. GET request handlers run *without* the engine lock (intentional — avoids blocking the UI during ticks). This meant a GET thread could be iterating the generator while the tick thread (which holds the lock) was simultaneously adding or removing instances (e.g. completing a project creates/removes `OngoingProject` entries).

**Fix:** Snapshot the dict at call time with `list(cls.instances().values())`. The caller iterates a stable copy rather than a live view. The return type changes from `Iterator[T]` to `list[T]`; all callers work with lists unchanged.

**Caveat:** `list()` itself still has a theoretical race window of microseconds (the C-level copy). In practice this is safe — the tick runs every 30s and the copy is near-instant. A 100% correct fix would require a per-dict lock, but that adds complexity without meaningful benefit given the observed frequency (2 occurrences).

### 2. OOM kill → corrupted save → crash loop

At 12:47 the kernel OOM-killed the process (547MB peak) mid-tick. Some `instance/data/` files had been written but `instance/engine_data.pck` had not been updated. On the next startup, `engine.load()` detects this mtime mismatch and aborts:

```
RuntimeError: The data has not been saved correctly, please restart form the last checkpoint.
```

The server then crash-looped (restarted 23+ times) because every restart hit the same check.

A clean service stop at 17:15 (while a tick was mid-write) re-triggered this, starting the second crash loop that lasted until manual intervention.

**Recovery:** Restore `checkpoints/last_checkpoint.tar.gz` while preserving `instance/actions_history.log`, which the engine replays on startup to catch up all player actions since the checkpoint. Without preserving the log, player actions since the last checkpoint are lost.

## Changes

- **`energetica/database/__init__.py`** — `DBModel.all()` now returns `list[T]` (snapshot) instead of a lazy `Iterator[T]`
- **`docs/backend/incident-recovery.md`** — new doc with the correct step-by-step recovery procedure, explanation of the mtime check, what the actions log is for, and a table of common causes

## Testing

`bun run typecheck` and `bun run lint` pass. The fix is a one-line change with no logic change beyond snapshotting.

<h3>Greptile Summary</h3>

This PR fixes the `RuntimeError: dictionary changed size during iteration` crash in `DBModel.all()` by snapshotting the instances dict with `list()` at call time, and adds a `touch()` call after `save_past_data()` to defend against coarse-mtime filesystem races that could trigger the data-integrity check on restart. It also fixes the checkpoint path (`last_checkpoint.tar.gz` → `checkpoints/last_checkpoint.tar.gz`) and adds an incident-recovery runbook.

- **`DBModel.all()` (database/__init__.py)**: Returns `list[T]` snapshot instead of a lazy generator; callers in `production_update.py` drop their now-redundant `list()` wrappers.
- **`tick_execution.py`**: `Path("instance/engine_data.pck").touch()` bumps the pickle's mtime after `save_past_data()`, ensuring it is strictly newer than all data files on filesystems with 1-second mtime resolution.
- **`energetica/__init__.py` + docs**: Checkpoint path corrected to `checkpoints/`; the `--load_checkpoint` handler correctly calls `shutil.rmtree("instance")` before extraction, so stale post-checkpoint data files cannot re-trigger the integrity check.

Safe to merge — the one-line snapshot fix is minimal and correct, the touch() workaround is well-scoped to the save_past_data() path, and the checkpoint path correction aligns the load path with the existing save path.

The core change is a single list() call that converts a lazy dict view into a stable snapshot, directly eliminating the observed crash. The touch() addition is a narrow, well-commented defensive measure confined to one branch. The checkpoint path fix corrects an existing inconsistency.

No files require special attention; energetica/utils/tick_execution.py has a minor comment clarity opportunity noted in the review.

| Filename | Overview |
|----------|----------|
| energetica/database/__init__.py | Core fix: `DBModel.all()` now returns `list[T]` snapshot instead of a lazy generator, eliminating the dict-changed-during-iteration race condition |
| energetica/utils/tick_execution.py | Adds `Path("instance/engine_data.pck").touch()` after `save_past_data()` to defend against coarse-mtime filesystem races; the `save_checkpoint()` path at end of tick doesn't get the same treatment but is safe because it writes a fresh pickle after all data files |
| energetica/__init__.py | Fixes checkpoint path from `last_checkpoint.tar.gz` (root) to `checkpoints/last_checkpoint.tar.gz`, consistent with `save_checkpoint()` in game_engine.py; also correctly uses `shutil.rmtree("instance")` before extraction, avoiding stale-file issues |
| energetica/production_update.py | Removes now-redundant `list()` wrappers around `Player.all()` and `Network.all()` since `DBModel.all()` already returns a list |
| docs/backend/incident-recovery.md | New runbook documenting crash-loop recovery, the integrity check mechanism, and the `--load_checkpoint` flag workflow; procedure relies on code-level `shutil.rmtree` for clean extraction, avoiding the stale-file tar issue |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant Tick as tick() [engine lock]
    participant SPD as save_past_data()
    participant FS as Filesystem
    participant GET as GET handler [no lock]

    Note over Tick,GET: Before fix — race condition
    Tick->>SPD: call (every 216 ticks)
    SPD->>FS: "write instance/data/**/*.pck"
    SPD->>FS: write instance/engine_data.pck
    GET->>+GET: DBModel.all() → lazy generator over live dict
    Tick->>Tick: check_events_completion() mutates dict
    GET-->>GET: RuntimeError: dict changed size during iteration

    Note over Tick,GET: After fix — snapshot
    GET->>GET: DBModel.all() → list(dict.values()) snapshot
    Tick->>Tick: mutates dict (safe, snapshot is stable)

    Note over Tick,FS: touch() guard for coarse-mtime FSes
    Tick->>SPD: call (every 216 ticks)
    SPD->>FS: "write data files (mtime = T)"
    SPD->>FS: "write engine_data.pck (mtime = T, same second)"
    Tick->>FS: "touch engine_data.pck (mtime = T' >= T)"
    Note over FS: engine.load() check: data <= pickle ✓
```

<sub>Reviews (4): Last reviewed commit: ["fix: remove redundant list() wrappers an..."](https://github.com/felixvonsamson/energetica/commit/ec3a4a6ee7416683cb543e5f5007b275899e6567) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=31210730)</sub>

### DIFF
--- a/docs/backend/incident-recovery.md
+++ b/docs/backend/incident-recovery.md
@@ -1,0 +1,90 @@
+# Incident Recovery
+
+## The data integrity check
+
+On startup, `engine.load()` (`game_engine.py:238`) compares the mtime of `instance/engine_data.pck` against the newest file in `instance/data/**/*`. If any data file is newer than the pickle, startup aborts with:
+
+```
+RuntimeError: The data has not been saved correctly, please restart form the last checkpoint.
+```
+
+This happens when the process is killed mid-save (OOM kill, SIGKILL, power loss) — some data files were written but `engine_data.pck` was never updated.
+
+## How recovery works
+
+`instance/actions_history.log` records every player action and every tick as a JSON line. On startup, after `engine.load()` succeeds, the engine reads this file and replays all actions recorded after the loaded tick. This means as long as the log is intact, no player actions are lost even when restoring an older checkpoint.
+
+**Never overwrite or delete `instance/actions_history.log` during recovery.**
+
+## Recovery procedure
+
+```bash
+ssh deploy@energetica-edu
+cd /var/www/energetica
+sudo systemctl stop energetica
+```
+
+**Step 1 — back up the current instance folder:**
+
+```bash
+sudo -u www-data cp -r instance/ instance.bak.$(date +%Y%m%d_%H%M%S)/
+```
+
+**Step 2 — preserve the actions log:**
+
+```bash
+sudo -u www-data cp instance/actions_history.log /tmp/actions_history.log.bak
+```
+
+**Step 3 — restore the checkpoint** (this overwrites `instance/` with a consistent snapshot):
+
+```bash
+sudo -u www-data tar -xzf checkpoints/last_checkpoint.tar.gz
+```
+
+**Step 4 — put the actions log back** (so the engine can replay all actions since the checkpoint):
+
+```bash
+sudo -u www-data cp /tmp/actions_history.log.bak instance/actions_history.log
+```
+
+**Step 5 — start the service:**
+
+```bash
+sudo systemctl start energetica
+sudo journalctl -u energetica -f
+```
+
+Watch the logs. You should see the engine replaying ticks rapidly until it catches up to the present, then resume normal 30-second ticks.
+
+## Verifying recovery
+
+Check the startup logs for a line like:
+
+```
+Loaded last checkpoint
+```
+
+followed by rapid tick replays (`t = XXXX`). If the engine logs an error instead, stop the service immediately before anything is overwritten, and investigate.
+
+## What can go wrong
+
+**The actions log is from a different instance than the checkpoint.** This happens if you restore the wrong checkpoint. The engine asserts `uuid` consistency and will abort — safe to retry with the correct checkpoint.
+
+**The checkpoint is very old and the actions log is huge.** Replay can take minutes. Don't interrupt it.
+
+**The actions log itself is corrupt or truncated.** The engine will abort on the malformed line. In this case you lose actions from the truncation point forward; restore without the log or replay up to the last valid tick.
+
+## Common causes
+
+| Symptom | Cause |
+|---|---|
+| OOM kill in journalctl (`result 'oom-kill'`) | Process exceeded memory limit mid-save |
+| `Deactivated successfully` just before crash loop | Process stopped cleanly mid-tick (manual stop, deploy) |
+| `dictionary changed size during iteration` errors before crash | Race condition between tick thread and request threads (separate bug) |
+
+## Checkpoints
+
+Checkpoints are saved to `checkpoints/last_checkpoint.tar.gz` by `save_checkpoint()` in `game_engine.py`. They are a tarball of the entire `instance/` directory including `engine_data.pck` and all data files at a consistent point in time.
+
+The checkpoint on `energetica-edu` is saved roughly every hour (check `ls -la checkpoints/` to see its age).
--- a/docs/backend/incident-recovery.md
+++ b/docs/backend/incident-recovery.md
@@ -19,40 +19,52 @@ This happens when the process is killed mid-save (OOM kill, SIGKILL, power loss)
 ## Recovery procedure
 
 ```bash
-ssh deploy@energetica-edu
+ssh root@energetica-edu
 cd /var/www/energetica
-sudo systemctl stop energetica
+systemctl stop energetica
 ```
 
 **Step 1 — back up the current instance folder:**
 
 ```bash
-sudo -u www-data cp -r instance/ instance.bak.$(date +%Y%m%d_%H%M%S)/
+cp -r instance/ instance.bak.$(date +%Y%m%d_%H%M%S)/
 ```
 
 **Step 2 — preserve the actions log:**
 
 ```bash
-sudo -u www-data cp instance/actions_history.log /tmp/actions_history.log.bak
+cp instance/actions_history.log /tmp/actions_history.log.bak
 ```
 
-**Step 3 — restore the checkpoint** (this overwrites `instance/` with a consistent snapshot):
+**Step 3 — remove the current instance folder** (prevents stale files with newer mtimes from surviving into the restored state):
 
 ```bash
-sudo -u www-data tar -xzf checkpoints/last_checkpoint.tar.gz
+rm -r instance/
 ```
 
-**Step 4 — put the actions log back** (so the engine can replay all actions since the checkpoint):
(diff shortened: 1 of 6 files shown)


docs/backend/incident-recovery.md
```diff
@@ -0,0 +1,90 @@
...
+
+```bash
+sudo -u www-data cp instance/actions_history.log /tmp/actions_history.log.bak
+```
+
+**Step 3 — restore the checkpoint** (this overwrites `instance/` with a consistent snapshot):
+
+```bash
+sudo -u www-data tar -xzf checkpoints/last_checkpoint.tar.gz
+```
```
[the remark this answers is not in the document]
felixvonsamson: That is a good point no ? 
We should probably add a `rm -r instance` before

docs/backend/incident-recovery.md
```diff
@@ -0,0 +1,90 @@
...
+```bash
+ssh deploy@energetica-edu
+cd /var/www/energetica
+sudo systemctl stop energetica
+```
+
+**Step 1 — back up the current instance folder:**
+
+```bash
+sudo -u www-data cp -r instance/ instance.bak.$(date +%Y%m%d_%H%M%S)/
```
felixvonsamson: Do not include `sudo -u www-data`

felixvonsamson: everywhere

yassir-akram:
last thing: touch engine.pck at the end, otherwise, server will detecht data inconstency

felixvonsamson:
Good catch @yassir-akram — fixed in 8237be51. `save_past_data()` writes to `instance/data/**/*` but `engine_data.pck` only gets updated later in the same tick by `engine.save()`. An OOM kill between those two is exactly what triggered today's crash loops (at t=27000, a multiple of 216). Now touching `engine_data.pck` immediately after `save_past_data()` closes that window.

felixvonsamson:
OK as you can tell I asked Claude to implement your remark but he did not do what I expected. Does the change that he suggest make sense @yassir-akram ?

mglst:
Right, I see:

> Step 6 — touch engine_data.pck (ensures its mtime is newer than all data files, preventing the integrity check from firing):

I would suggest maybe making this be... a command line flag for main. Using `touch` to 'refresh' the mtime of the `engine_data.pck` is rather smart, but also rather a workaround.
The procedure would then instead be something like:

> Step 6: call main.py but with a special flag to get it to resimulate from action history, but without servering APIs and without making use of `systemd` (which is step 7)

mglst:
I'm a fan of `docs/backend/incident-recovery.md`! Good stuff. It's been sorely missing so far.

felixvonsamson:
there is a build-in recovery procedure in main made by @yassir-akram.

mglst:
Right, the `--simulate_file` flag.

Okay, then let's update `docs/backend/incident-recovery.md` to make use of it then in step 6

docs/backend/incident-recovery.md
```diff
@@ -0,0 +1,102 @@
...
+tar -xzf checkpoints/last_checkpoint.tar.gz
+```
+
+**Step 5 — put the actions log back** (so the engine can replay all actions since the checkpoint):
+
+```bash
+cp /tmp/actions_history.log.bak instance/actions_history.log
+```
+
+**Step 6 — touch `engine_data.pck`** (ensures its mtime is newer than all data files, preventing the integrity check from firing):
```
mglst: `--simulate_file` flag

mglst (changes requested):
Document the `--simulate_file` in step 6

energetica/database/__init__.py
```diff
@@ -81,9 +81,9 @@ def getitem(cls: type[T], id: int, error: Exception | None = None) -> T:  # pyli
             raise error
 
     @classmethod
-    def all(cls: type[T]) -> Iterator[T]:
+    def all(cls: type[T]) -> list[T]:
         """Get all instances of this class."""
-        return (v for v in cls.instances().values())
+        return list(cls.instances().values())
```
mglst, line 86: There are now places, e.g. `production_update.py:47`, that double-wrap with `list()`

Harmless, but noise.

We'll revisit this once we go back to using SQL.

docs/backend/incident-recovery.md
felixvonsamson: Actually we should use `--load_checkpoint`

energetica/utils/tick_execution.py
```diff
@@ -38,6 +39,7 @@ def tick() -> None:
     engine.log(f"t = {engine.total_t}")
     if engine.total_t % 216 == 0:
         save_past_data()
+        Path("instance/engine_data.pck").touch()
```
mglst, line 45: ideally this line would have a comment explaining the subtlety inline. But fine!

mglst (approved):

mglst:
LGTM.

@felixvonsamson I'll let you resolve all outstanding comments - I'm happy to consider them resolved as is.

PR #774, opened 2026-05-10
adding script to extract game instance data to csv

felixvonsamson:
<h3>Greptile Summary</h3>

This PR adds a new standalone script that reads a game instance's pickle files and exports time-series data to CSV across four resolutions (1d / 6d / 36d / 216d).

- `export_players` and `export_networks` iterate over all players/networks, skipping missing files and writing one CSV per resolution.
- `export_market_last` locates the highest-tick chart file numerically (via `max` + `int` extraction), then flattens supply/demand bid curves into a single CSV.
- `export_climate` checks for the climate pickle before opening it, matching the guard pattern used in every other export function.

Safe to merge — the script is a read-only export utility that writes CSVs and has no effect on the running game state.

All three issues surfaced in earlier review rounds (lexicographic sort, empty-directory crash, missing climate-file guard) are now addressed. The remaining code is straightforward pickle-to-CSV mapping with per-file existence checks throughout.

| Filename | Overview |
|----------|----------|
| scripts/export_instance_to_csv.py | New utility script that exports game instance pickle data to CSV files across four time resolutions; guarded file-existence checks for all data sources, numeric sort for chart files, and empty-file warnings are all correctly in place. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A[main] --> B[Load engine_data.pck]
    B --> C{exists?}
    C -- No --> Z[sys.exit 1]
    C -- Yes --> D[Build id_to_username map]
    D --> E[export_players]
    D --> F[export_networks]
    D --> G[export_market_last]
    D --> H[export_climate]

    E --> E1[For each RESOLUTION]
    E1 --> E2[time_points]
    E2 --> E3[For each player]
    E3 --> E4{player_N.pck exists?}
    E4 -- No --> E3
    E4 -- Yes --> E5[Extract revenues / generation / demand / storage / emissions]
    E5 --> E6[players_name.csv]

    F --> F1[For each RESOLUTION]
    F1 --> F2[time_points]
    F2 --> F3[For each network]
    F3 --> F4{time_series.pck exists?}
    F4 -- No --> F3
    F4 -- Yes --> F5[Extract price / quantity / exports / imports / generation]
    F5 --> F6[network_name.csv]

    G --> G1[For each network]
    G1 --> G2{charts dir exists?}
    G2 -- No --> G1
    G2 -- Yes --> G3{market_t files found?}
    G3 -- No --> G1
    G3 -- Yes --> G4[max by tick number]
    G4 --> G5[Extract supply & demand curves]
    G5 --> G6[market_last.csv]

    H --> H1{climate_data.pck exists?}
    H1 -- No --> H2[warn & return]
    H1 -- Yes --> H3[For each RESOLUTION]
    H3 --> H4[time_points]
    H4 --> H5[Extract emissions / temperature]
    H5 --> H6[climate_name.csv]
```

<sub>Reviews (2): Last reviewed commit: ["fix: address Greptile review issues in e..."](https://github.com/felixvonsamson/energetica/commit/248ccf7e2f2252e70fd993c61fd25107ef960a4e) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=31488045)</sub>

### DIFF
--- a/scripts/export_instance_to_csv.py
+++ b/scripts/export_instance_to_csv.py
@@ -162,7 +162,11 @@ def export_market_last(base, out, eng, id_to_username):
         charts_dir = f"{base}/data/networks/{net_id}/charts"
         if not os.path.exists(charts_dir):
             continue
-        last_file = sorted(os.listdir(charts_dir))[-1]
+        files = [f for f in os.listdir(charts_dir) if re.search(r"market_t(\d+)", f)]
+        if not files:
+            print(f"  Warning: no chart files found in {charts_dir}, skipping network {net_id}")
+            continue
+        last_file = max(files, key=lambda f: int(re.search(r"market_t(\d+)", f).group(1)))
         tick_num = int(re.search(r"market_t(\d+)", last_file).group(1))
 
         with open(f"{charts_dir}/{last_file}", "rb") as f:
@@ -196,7 +200,11 @@ def export_climate(base, out, eng):
     clock_time = eng["clock_time"]
     start_date = eng["start_date"]
 
-    with open(f"{base}/data/servers/climate_data.pck", "rb") as f:
+    climate_path = f"{base}/data/servers/climate_data.pck"
+    if not os.path.exists(climate_path):
+        print(f"Warning: {climate_path} not found, skipping climate export.", file=sys.stderr)
+        return
+    with open(climate_path, "rb") as f:
         climate = pickle.load(f)
 
     for res_idx, factor, name in RESOLUTIONS:
--- a/scripts/export_instance_to_csv.py
+++ b/scripts/export_instance_to_csv.py
@@ -1,0 +1,253 @@
+"""
+Export an Energetica instance snapshot to CSV files.
+
+Usage (run from the repo root):
+    python scripts/export_instance_to_csv.py <instance_dir> <output_dir>
+
+Defaults:
+    instance_dir : instance/
+    output_dir   : energetica_export/
+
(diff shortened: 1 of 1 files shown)


mglst:
It looks like there are greptile comments worth addressing - I'll hold off on reviewing thoroughly for a bit

mglst:
Hmm. I was thinking "how about we make this into an endpoint?" but then actually this is a rather 'expensive' operation which would bog down the server, so we probably _don't_ want it to be an API.

felixvonsamson:
@greptile-apps Thank you SO much for the thorough review — genuinely one of the most useful pieces of automated feedback I've received on this project! 🙏

The lexicographic sort bug was a particularly sneaky one: it would have silently produced wrong market data for any real game instance past tick 9, and I might never have noticed without the explicit call-out. The empty-directory guard and the missing existence check in `export_climate` are also exactly the kind of edge cases that are easy to overlook when writing a quick utility script.

All three points have been addressed in the follow-up commit. Really appreciate the clear explanations, the severity labels, and the concrete code suggestions — made it very easy to act on. Keep up the excellent work! 🎉

mglst (approved):
Nice. The top level 'shape' / interaction modes of the script seem good. The one thing that jumps out at me is that the market data is only for the latest tick. I suppose that's fine actually - we don't store that data for very long either way, so a single snapshot is really the best we can expect. Good. I didn't look at the actual implementation, but I trust that it works as described.

PR #782, opened 2026-05-15
Prevent storage price inversions in priority list

### LINKED ISSUE #306: Prevent nonsensical reorder / pricing of storage facility charge / discharge
It doesn't make sense to have the wrong order the charge / discharge of a storage facility or to set the prices in a network to the wrong order. I suggest we add some feedback to warn or prevent this

I suggest that we implement the following:
* either players are simply forbidden to make these actions
* or the players are allowed to do these actions, but they are given a warning

felixvonsamson:
Fixes #306.

## Summary

- **Backend validation**: `NetworkPrices.update()` now checks that no storage facility ends up with a discharge (ask) price lower than its charge (bid) price before applying any price change. Merging current prices with the incoming update before checking correctly handles partial updates (e.g. only changing the ask side). Raises a new `storagePriceInversion` error, which surfaces as a toast.
- **Bug fix**: `validateBatteryOrder` had its condition inverted — it was marking the valid default state (discharge price ~800, charge price ~200) as invalid. Fixed.
- **UX fix**: `PriceInput` now awaits the async `onCommit` and resets to the last valid value if the mutation rejects, instead of leaving the rejected value in the field until a page refresh.

## Test plan

- [ ] Try moving a storage discharge row above its charge row via bump buttons → toast appears, order unchanged
- [ ] Try typing a discharge price lower than the current charge price → toast appears, input reverts to previous value
- [ ] Try typing a charge price higher than the current discharge price → same
- [ ] Valid price changes and reorders still work normally

<h3>Greptile Summary</h3>

This PR closes #306 by moving storage price-inversion enforcement from a broken frontend utility to a correct backend validation in `NetworkPrices.update()`, and fixes the `PriceInput` component to revert on rejected mutations.

- **Backend guard**: `NetworkPrices.update()` merges current prices with the incoming partial update before checking every `StorageFacilityType`; raises `STORAGE_PRICE_INVERSION` if any discharge (ask) price would end up below its charge (bid) price. Both reorder bumps and direct price edits flow through `update()`, so the constraint is enforced uniformly.
- **Bug fix**: `validateBatteryOrder` in `power-priorities-utils.ts` had its condition inverted — `dischargeIdx > chargeIdx` is the *valid* state in an ascending-price list, but the function returned `false` for it. The function is removed entirely; the server is now the single source of truth.
- **UX fix**: `PriceInput.commit` now `await`s the async `onCommit` and resets `localValue` to the last server-provided `value` when the promise rejects, so a rejected price change no longer persists in the field until a page refresh.

Safe to merge — the change is a pure addition of a new constraint that prevents an invalid game state, with no modifications to existing happy-path logic.

The backend guard is atomic (validation happens before mutation), the error propagation chain from Python through FastAPI to the React toast and field-reset is complete and correct, and the removed frontend function had an inverted condition so its deletion is strictly an improvement. No existing paths are affected when prices are valid.

| Filename | Overview |
|----------|----------|
| energetica/database/network_prices.py | Adds storage price inversion guard to `update()` using merged ask/bid dicts so partial updates are validated correctly; `change_facility_priority` inherits the check by calling `update()` |
| energetica/game_error.py | Adds `STORAGE_PRICE_INVERSION = "storagePriceInversion"` enum value; strings match the frontend message key and generated API type exactly |
| frontend/src/components/power-priorities/price-input.tsx | Makes `onCommit` async and wraps the call in try/catch; on rejection resets localValue to the server-provided `value` prop — correct because value hasn't changed if the mutation was rejected |
| frontend/src/lib/game-messages.ts | Adds user-facing message for the new error key |
| frontend/src/lib/power-priorities-utils.ts | Removes `validateBatteryOrder`, which had an inverted condition (was treating the valid state as invalid); validation is now fully server-side |
| frontend/src/types/api.generated.ts | Adds `storagePriceInversion` to the generated `GameExceptionType` union, keeping backend and frontend error-type contracts in sync |

</details>

<sub>Reviews (3): Last reviewed commit: ["Removed dead code"](https://github.com/felixvonsamson/energetica/commit/21b67ff1cc478bfdcfd9e44bc10879ea2b856126) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=32374645)</sub>

### DIFF
--- a/frontend/src/lib/power-priorities-utils.ts
+++ b/frontend/src/lib/power-priorities-utils.ts
@@ -13,11 +13,15 @@ const STORAGE_FACILITY_TYPES = [
 ] as const;
 
 /**
- * Validates that battery discharge (ask) is not placed below battery charge
- * (bid) in the unified priority list.
+ * Validates that no storage facility has its discharge price below its charge
+ * price in the unified priority list.
+ *
+ * In the unified list, lower index = lower price. A discharge (ask) item placed
+ * before a charge (bid) item of the same type would receive a lower price,
+ * meaning the facility sells for less than it buys — an economic loss.
  *
  * @param priorities - The unified priority array
- * @returns True if valid, false if any battery has discharge below charge
+ * @returns True if valid, false if any storage type has discharge before charge
  */
 export function validateBatteryOrder(priorities: PowerPriorityItem[]): boolean {
     for (const storageType of STORAGE_FACILITY_TYPES) {
@@ -28,12 +32,12 @@ export function validateBatteryOrder(priorities: PowerPriorityItem[]): boolean {
             (p) => p.type === storageType && p.side === "bid",
         );
 
-        // Invalid: discharge (ask) must not be AFTER charge (bid) in priority list
-        // Lower index = higher priority
+        // Invalid: discharge (ask) must not be BEFORE charge (bid) in priority list,
+        // as that would assign a lower sell price than buy price.
         if (
             dischargeIdx !== -1 &&
             chargeIdx !== -1 &&
-            dischargeIdx > chargeIdx
+            dischargeIdx < chargeIdx
         ) {
             return false;
         }
--- a/frontend/src/lib/power-priorities-utils.ts
+++ b/frontend/src/lib/power-priorities-utils.ts
@@ -13,40 +13,6 @@ const STORAGE_FACILITY_TYPES = [
 ] as const;
 
 /**
- * Validates that no storage facility has its discharge price below its charge
- * price in the unified priority list.
- *
- * In the unified list, lower index = lower price. A discharge (ask) item placed
- * before a charge (bid) item of the same type would receive a lower price,
- * meaning the facility sells for less than it buys — an economic loss.
- *
- * @param priorities - The unified priority array
- * @returns True if valid, false if any storage type has discharge before charge
- */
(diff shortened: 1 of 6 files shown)


frontend/src/lib/power-priorities-utils.ts
```diff
@@ -13,11 +13,15 @@ const STORAGE_FACILITY_TYPES = [
...
+ *
+ * In the unified list, lower index = lower price. A discharge (ask) item placed
+ * before a charge (bid) item of the same type would receive a lower price,
+ * meaning the facility sells for less than it buys — an economic loss.
  *
  * @param priorities - The unified priority array
- * @returns True if valid, false if any battery has discharge below charge
+ * @returns True if valid, false if any storage type has discharge before charge
  */
 export function validateBatteryOrder(priorities: PowerPriorityItem[]): boolean {
```
felixvonsamson: Is never called. @mglst is that normal ?

mglst: no, was probably leftover dead code that never got cleaned up. Cleared to remove.

mglst (approved):
Clean. Largely just backend though. Would be nice for the frontend to also show better visual treatment of forbidden actions rather than waiting for the backend to complain. But I'm nitpicking. Ship it.

Issue #793, opened 2026-05-22
Apply Top-N + Other bucketing uniformly to player-breakdown charts


mglst:
## Context

The "Clearing Volume by Player Exports/Imports" charts on `/app/overviews/electricity-markets` (and any future player-breakdown charts, including the upcoming Market Flows Sankey) currently render one series per player. In markets with many participants, this produces:

- Unreadable legends and overlapping labels
- Long-tail slivers that contribute visual noise without information
- Inconsistent legibility across markets of different sizes

## Proposal

Apply a uniform Top-N + "Other" bucketing strategy to all player-breakdown chart data:

- Sort participants by cleared MW (or relevant metric) descending
- Keep the top N (suggested N=8)
- Sum the remaining into a single `Other (N players)` bucket per side
- "Other" gets a neutral muted color
- Tooltip on "Other" lists rolled-up names + individual contributions

Total link/area widths must continue to sum to the clearing volume (no silent drops).

## Affected surfaces (non-exhaustive)

- `frontend/src/components/charts/market-clearing-volume-chart.tsx` — player breakdown mode
- `frontend/src/components/charts/market-flow-sankey.tsx` — to be added, see #TBD (Sankey prototype)
- Any future player-keyed breakdown views

## Non-goals

- Not user-configurable; a sensible N is better than another picker
- Type-breakdown views (~10 facility types max) don't need this; only player-breakdown views

## Origin

Surfaced during design grilling for the Market Flows Sankey prototype, where we decided to ship the prototype without bucketing (acceptable for an experiment) and file this for uniform follow-up across all player-breakdown charts.

felixvonsamson:
I suggest doing the same for the production by technology. maybe rather something like : 

> If a technology/player produced/consumed less than 1% of the energy in the displayed timeframe, it is listed under others. 

Rather than a fixed top-N.

PR #794, opened 2026-05-22
feat: Market Flows Sankey on electricity-markets overview

mglst:
## Summary

Adds a 3-column Sankey (sellers → Market → buyers) alongside the existing Merit Order chart on `/app/overviews/electricity-markets`. Center node is labeled with cleared volume and price. Sources and sinks group by player or facility type via a shared Color-by control.

Restructures the page to share snapshot state between the Merit Order chart and the new Sankey:
- Lifts the tick slider and Color-by picker into a **sticky** page-level *Snapshot controls* PageCard.
- Removes them from `MeritOrderChart` (now takes `selectedTick` and `colorMode` as props; `activeMode` stays local).
- Wraps the snapshot section in a flex container so sticky controls pin while scrolling between the two charts, then release once the section exits the viewport.

Reuses the existing breakdown chart-data endpoints (`market-exports` / `market-imports` / `market-generation` / `market-consumption`) which already return per-tick pre-aggregated values, so no client-side merit-order walking is needed and slider scrubbing is in-memory free (no refetch).

## Design choices (from grilling session)

- **ECharts `SankeyChart`** rather than Mermaid — integrates with existing color tokens (`useAssetColorGetter`, `getHashBasedChartColor`), dark mode, tooltips, and ResizeObserver patterns. Zero new deps.
- **3 columns with explicit center *Market* node** rather than 2-column all-to-all — we have two independent breakdowns, not seller→buyer pairings. The center node honestly represents pool clearing and provides a natural slot for the volume + price callout.
- **Label suppression** at `<2%` share rather than Top-N + Other bucketing (deferred uniformly across player-breakdown charts in #793).
- **`draggable: false`** — node drag adds no value for a snapshot view (state resets on slider scrub) and is awkward on touch.
- **Hover blur softened** — ECharts' default emphasis blur drops non-adjacent items to ~0.1 opacity (reads as flicker); overridden to 0.6 / 0.3 so non-adjacent items stay legible.
- **Sink labels positioned `left`**, market label positioned `top` — prevents right-side overflow and avoids crowding the link fanout.

## Status

Prototype / experiment. Worth keeping behind eyes-on review before relying on this in any educational flow.

## Test plan

- [ ] Open `/app/overviews/electricity-markets` while signed in as a player whose market has multiple participants
- [ ] Scrub the slider, verify Merit Order **and** Sankey both update to the same tick
- [ ] Toggle Color-by: Player ↔ Technology — both charts should re-color in lockstep
- [ ] Scroll between Merit Order and Sankey — controls card should stay pinned within the snapshot section, release once scrolled past
- [ ] Switch markets via the selector — slider should reset to "Current"
- [ ] Hover Sankey nodes / links — tooltip shows display name + MW + share; non-adjacent items dim but stay legible (no flicker)
- [ ] Right-side (sink) labels render inside the chart, not cut off
- [ ] Sankey nodes are not draggable
- [ ] Empty markets or pre-creation ticks render "No flows for this tick" rather than crashing

## Related

- #793 — Apply Top-N + Other bucketing uniformly to player-breakdown charts

### DIFF
--- a/frontend/src/components/charts/market-flow-sankey.tsx
+++ b/frontend/src/components/charts/market-flow-sankey.tsx
@@ -1,0 +1,403 @@
+/**
+ * Sankey diagram of electricity market flows at a single tick.
+ *
+ * Visualizes sellers → Market → buyers with link widths proportional to cleared
+ * MW. Sources and sinks are grouped either by player or by facility type,
+ * controlled by `colorMode`. The center node is labeled with the clearing
+ * volume and price.
+ *
+ * Reuses the breakdown chart endpoints (market-exports / market-imports /
+ * market-generation / market-consumption) which already return per-tick
+ * pre-aggregated values, so no client-side merit-order walking is needed.
+ */
+
+import { SankeyChart } from "echarts/charts";
+import type { SankeySeriesOption } from "echarts/charts";
+import { TooltipComponent } from "echarts/components";
+import type { TooltipComponentOption } from "echarts/components";
+import type { ComposeOption } from "echarts/core";
+import * as echarts from "echarts/core";
+import { CanvasRenderer } from "echarts/renderers";
+import { memo, useEffect, useMemo, useRef } from "react";
+
+import type { ColorMode } from "@/components/charts/supply-demand-chart";
+import { useAssetColorGetter } from "@/hooks/use-asset-color-getter";
+import { useChartData } from "@/hooks/use-charts";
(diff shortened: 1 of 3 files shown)


mglst:
"neat but not immediately useful" - Felix
I suggested could be used for admin views or for instance summary once finished. "Sounds good" - Felix
So keep as draft / close, and revisit when we get around to these.

PR #803, opened 2026-06-11
rfc: phase 2 — server-wide accounts

mglst:
## RFC Phase 2 — Server-Wide Accounts

Second phase of the [Apache Static Serving, Split Build & Multi-Instance Deployment RFC](../blob/rfc/phase-2-server-wide-accounts/docs/architecture/static-serving-and-deployment.md). Introduces a server-wide SQLite identity store shared across all instances on a VPS, decoupling credentials from per-instance engine pickles. Builds on Phase 1 (#791).

### What changed
- **`energetica/accounts/`** — SQLite store: WAL mode, lazy schema bootstrap, `ENERGETICA_ACCOUNTS_DB_PATH` env var (default `/var/lib/energetica/accounts.db`). Connection-per-call.
- **Pickle `User` gains required `account_id`** (FK to `accounts.account_id`). `__setstate__` raises a clear error if missing, forcing the migration to run before new code starts.
- **Auth router**
  - Signup: SQLite-row-first + pickle-row-second, with rollback (`accounts.delete_account`) if the pickle write fails.
  - Login: auto-provisions a pickle `User(account_id, role="player", player=None)` on first visit to an instance. Full access-policy gating is deferred to Phase 3 — current behaviour treats every instance as `public`.
  - Password change writes through to the accounts store.
- **`scripts/migrate-to-server-accounts.py`** — idempotent pickle→SQLite backfill (`INSERT OR IGNORE` + `SELECT`), `--dry-run` supported.
- **Tests** — split `test_auth_routes.py` into `test_login_route` / `test_signup_route` / `test_change_password_route`; add `test_accounts_db` and `test_user_account_id` unit tests; add `conftest.py` fixtures.

### ⚠️ Deployment ordering (required)
This phase changes both the **auth path** and the **pickle shape**. The migration script **must run on each VPS before the new code starts**:

1. Stop the instance service.
2. Create the accounts dir: `sudo mkdir -p /var/lib/energetica && sudo chown energetica:energetica /var/lib/energetica`
3. Run migration as the service user: `sudo -u energetica .venv/bin/python scripts/migrate-to-server-accounts.py --pickle <engine_data.pck>` (use `--dry-run` first).
4. Deploy new code, restart service.

If new code lands first, every login raises `RuntimeError: Pickle User is missing account_id …`. Recovery: stop service, run migration, restart. Full notes in the RFC doc § *Phase 2 deployment notes*.

### Testing
[...]

PR #807, opened 2026-06-15
rfc: phase 3 — instance visibility & access policy

mglst:
Phase 3 of the static-serving / multi-instance RFC (`docs/architecture/static-serving-and-deployment.md` § Instance Visibility & Access). Builds on Phase 2 (#803).

## What this does
Reads each instance's access policy from `instance.json` fresh on every login/signup, gates accordingly, and publishes a sanitised public fragment + aggregate manifest for the landing page.

### Backend
- **`energetica/instance_config.py`** — reads `{ENERGETICA_INSTANCE_CONFIG_DIR}/{slug}/instance.json` (default `/etc/energetica`) **fresh per attempt, no cache**, so admin policy edits take effect on the next login with no restart. Slug + dirs are env-configurable for hermetic tests, mirroring the Phase-2 accounts-DB convention.
- **`auth.py`** — `login` **and** `signup` enforce the policy *after* the credential check: `public`/unconfigured allows any server-wide account; `private` allows only `allowed_usernames`. Denied → `403 INSTANCE_ACCESS_DENIED`. Gating signup too closes the hole where a private/unadvertised instance reachable by URL could otherwise mint server-wide accounts.
- On startup and on each allowed login, publishes the sanitised fragment (`access` block stripped) to `{landing}/instances/{slug}.json` and re-aggregates `{landing}/instances.json` (sorted `starts_at` desc). Atomic write via unique `mkstemp` tmp + `os.replace`; **best-effort — a publish failure never breaks login**.

### Frontend (landing bundle)
- `useAdvertisedInstances` + `AdvertisedRuns` fetch `/instances.json` and render advertised runs; the signup CTA targets each run's `/app/sign-up` (cross-subdomain via `VITE_APEX_DOMAIN`, same-origin relative in the interim). Degrades to nothing before Apache serves the manifest (Phase 4/5).

## Decisions worth a reviewer's eye
- **Aggregation is pure Python, not a `jq` subprocess** — identical `{instances: [...]}` output, testable without `jq`, no subprocess on the login path. (`jq` is still installed by `setup-base.sh` for the other infra scripts.)
- **Security defaults:** absent `instance.json` ⇒ `public` (preserves Phase-2/dev behaviour); present-but-unparseable ⇒ **fail closed (403)** — including a half-edited config with a stray key in the `access` block (`extra="forbid"`). `starts_at` must be tz-aware (naive ⇒ fail closed), which also keeps the aggregation sort from mixing naive/aware datetimes.
- **`api.generated.ts`** is regenerated: beyond the new `INSTANCE_ACCESS_DENIED` enum it also picks up prior-phase backend drift (removed `/admin-dashboard`, `/sign-up` redirect, change-password docstring) that had not been regenerated.

## Known limitation (private instances)
Access is gated purely on `allowed_usernames` with no role bypass, so the auto-provisioned `admin` account is locked out of a `private` instance unless explicitly listed. Acceptable while the focus is public instances; private-instance administration is one of the design decisions deferred with the rest of the private/unadvertised work. The existing global `disable_signups` + `players.txt` mechanism also overlaps with the allowlist; they compose for now (the `disable_signups` check runs first), unification deferred (migration implications).

## Tests
Unit: load/validate matrix incl. fail-closed cases (corrupt, unknown policy, extra key, naive datetime, half-edited public+allowlist), allow/deny incl. empty allowlist, fragment strips `access`, aggregation ordering. Integration: public/private login & signup, denied-no-provision, lock-down-after-provision, fail-closed-on-corrupt, wrong-password-is-401-not-403. Full suite green (minus the two pre-existing known-unrelated failures).

<h3>Greptile Summary</h3>

Implements per-instance visibility and access-policy gating (Phase 3 of the static-serving RFC). `instance.json` is read fresh on every login/signup (no in-memory cache), enabling zero-restart policy updates; absent or unparseable files fail closed with `403 INSTANCE_ACCESS_DENIED`.

- **Backend**: `instance_config.py` handles config loading, public/private discrimination, and atomic POSIX-safe writes of per-instance fragments and the aggregate manifest; `auth.py` gates both login and signup after the credential check.
- **Publish pipeline**: on each allowed login (and startup), a sanitised `InstanceFragment` (access block stripped) is atomically written to the landing dir; `aggregate_instances()` re-aggregates `instances.json` filtering to `advertised: true` entries only, preventing subdomain enumeration of hidden instances.
- **Frontend**: `useAdvertisedInstances` + `AdvertisedRuns` fetch `/instances.json` and render run cards linking to each instance signup; degrades gracefully before Phase 4/5 Apache wiring.

Safe to merge. The access-gating and publish pipeline are well-guarded: corrupt configs fail closed, unadvertised instances are excluded from the public manifest, atomic writes prevent partial reads, and the fd-ownership flag prevents descriptor leaks.

The core logic is straightforward and thoroughly tested at both unit and integration levels. The fail-closed design means any misconfiguration causes a 403 rather than silent open access. The publish path is best-effort and can never break the login flow.

No files require special attention. The only nit is a redundant advertised filter in use-instances.ts.

| Filename | Overview |
|----------|----------|
| energetica/instance_config.py | Core of the PR: reads instance.json per-request, enforces public/private policy, atomically publishes sanitised fragments and the aggregate manifest. Fail-closed on corrupt config; unadvertised instances correctly excluded from the public manifest; fd-leak fix included. |
| energetica/routers/auth.py | Adds _enforce_instance_access() called after credential checks in both login and signup. Deny paths never reach publish(); best-effort publish only runs on allowed access. Ordering is intentional and tested. |
| tests/integration/test_instance_access.py | Good integration coverage: public/private login and signup, denied-no-provision, lock-down-after-provision, fail-closed-on-corrupt, wrong-password-returns-401-not-403, fragment publication verified. |
| tests/unit/test_instance_config.py | Comprehensive unit tests: load/validate matrix, fail-closed cases, allow/deny, empty allowlist, fragment strips access, aggregation ordering, unadvertised exclusion. |
| frontend/src/hooks/use-instances.ts | Fetches the public manifest and filters for advertised instances. The frontend filter is redundant (backend already emits only advertised entries), but harmless. AbortController cleanup is correctly implemented. |
| frontend/src/lib/instances.ts | Defines InstanceFragment type and instanceSignupHref; cross-origin subdomain links when VITE_APEX_DOMAIN is set, same-origin fallback otherwise (documented known limitation). |

</details>

<h3>Sequence Diagram</h3>

<a href="#gh-light-mode-only">

```mermaid
sequenceDiagram
    participant Client
    participant Auth as auth.py
    participant IC as instance_config.py
    participant FS as Landing FS

    Client->>Auth: POST /auth/login (username, password)
    Auth->>Auth: Check account exists
    Auth->>Auth: Verify password hash
    Auth->>IC: _enforce_instance_access(username)
    IC->>FS: Read instance.json (fresh, no cache)
    alt File absent or slug unset
        IC-->>Auth: "config = None (public)"
    else File valid
        IC-->>Auth: InstanceConfig
    else File corrupt
        Auth-->>Client: 403 INSTANCE_ACCESS_DENIED
    end
    alt private and username not in allowlist
        Auth-->>Client: 403 INSTANCE_ACCESS_DENIED
    else Allowed
        IC->>FS: Write sanitised fragment + re-aggregate instances.json
        Auth-->>Client: 200 OK + session cookie
    end
```

</a>
<a href="#gh-dark-mode-only">

```mermaid
%%{init: {'theme': 'base', 'themeVariables': {"darkMode": true, "background": "#0d1117", "primaryColor": "#21262d", "primaryTextColor": "#e6edf3", "primaryBorderColor": "#8b949e", "lineColor": "#8b949e", "textColor": "#e6edf3", "edgeLabelBackground": "#161b22", "actorBkg": "#21262d", "actorBorder": "#8b949e", "actorTextColor": "#e6edf3", "actorLineColor": "#8b949e", "signalColor": "#8b949e", "signalTextColor": "#e6edf3", "noteBkgColor": "#373320", "noteBorderColor": "#d4a72c", "noteTextColor": "#f0e6c0", "labelBoxBkgColor": "#21262d", "labelBoxBorderColor": "#8b949e", "labelTextColor": "#e6edf3", "loopTextColor": "#e6edf3", "activationBkgColor": "#30363d", "activationBorderColor": "#8b949e"}}}%%
sequenceDiagram
    participant Client
    participant Auth as auth.py
    participant IC as instance_config.py
    participant FS as Landing FS

    Client->>Auth: POST /auth/login (username, password)
    Auth->>Auth: Check account exists
    Auth->>Auth: Verify password hash
    Auth->>IC: _enforce_instance_access(username)
    IC->>FS: Read instance.json (fresh, no cache)
    alt File absent or slug unset
        IC-->>Auth: "config = None (public)"
    else File valid
        IC-->>Auth: InstanceConfig
    else File corrupt
        Auth-->>Client: 403 INSTANCE_ACCESS_DENIED
    end
    alt private and username not in allowlist
        Auth-->>Client: 403 INSTANCE_ACCESS_DENIED
    else Allowed
        IC->>FS: Write sanitised fragment + re-aggregate instances.json
        Auth-->>Client: 200 OK + session cookie
    end
```

</a>

<sub>Reviews (3): Last reviewed commit: ["review: publish instance manifests group..."](https://github.com/felixvonsamson/energetica/commit/ab5121cdd2e1a8611ab70f6331f66a54e29be2a2) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=37718114)</sub>


mglst:
Thanks @greptile — both addressed in cdc62443:

- **P1 (unadvertised slugs leaking into `instances.json`):** `aggregate_instances()` now filters to `advertised: true` before writing the manifest, so an unadvertised instance keeps its on-disk fragment (reachable only if you already know the slug) but is no longer enumerable from the world-readable `instances.json`. Added `test_aggregate_excludes_unadvertised_instances` and updated the RFC (the original design relied on the frontend filtering; moved server-side for defence in depth).
- **P2 (fd leak on `os.fdopen` failure):** `_atomic_write_json` now tracks fd ownership with a flag and only `os.close(fd)` when `fdopen` never took ownership — a blind close in the `except` would risk a thread-unsafe double-close on a recycled fd.

mglst:
## Fragment readability: published manifests are unreadable by Apache (surfaced during Phase 4)

While building the Phase 4 Apache/infra layer I hit a contract gap that originates here and should be fixed on this PR rather than left broken on `main` until Phase 4 lands.

### The problem
`_atomic_write_json` (`energetica/instance_config.py:147`) creates the temp file with `tempfile.mkstemp` (mode **`0600`**, `instance_config.py:157`) and then `os.replace(tmp_name, target)` (`:163`). `rename(2)` preserves the source inode's mode, so every published file ends up **`0600`, owned by the instance's service user**:

- `/var/www/energetica-landing/instances/{slug}.json`
- `/var/www/energetica-landing/instances.json`

In Phase 4+ Apache (`www-data`) serves the landing dir statically. It is **not** the file owner, so a `0600` file is unreadable → the instance picker manifest and per-instance fragments 403/fail to load. The entire publish→aggregate→serve contract this PR builds is inert once Apache fronts the landing.

### Proposed fix (one line, here)
Make the published files group-readable before the atomic rename:

```python
        with os.fdopen(fd, "w", encoding="utf-8") as tmp_file:
            fd_open = False
            tmp_file.write(payload)
        os.chmod(tmp_name, 0o640)   # mkstemp forces 0600; Apache (www-data) must read these
        os.replace(tmp_name, target)
```

`0640` (group read), **not** `0644` — Phase 4 adds `www-data` to a shared `energetica` group and the landing dir is `setgid energetica`, so group-read is sufficient and keeps the files from being world-readable on disk (defense in depth; the HTTP exposure is intentional, the filesystem one need not be).

### Why land it here
If this ships only in Phase 4, `main` spends the gap between PRs publishing manifests no web server can read. One line on this PR keeps every phase independently deployable (the stated invariant of this RFC).

Phase 4 infra is staged on `rfc/phase-4-apache-infra` and already (a) puts `www-data` in the `energetica` group and (b) documents this in the RFC status block.

@greptileai — please review this proposed change; flagging in case there's a reason the fragments were intentionally `0600` that I'm missing.

mglst:
Landed as `ab5121cd` — `os.chmod(tmp_name, 0o640)` before the atomic rename in `_atomic_write_json`, plus `test_publish_files_are_group_readable` asserting both the per-instance fragment and the aggregate `instances.json` come out `0o640`. Single code path, so both are covered. Thanks for the confirm @greptileai.

felixvonsamson (approved):
Did not review the code in detail but looks good. 
A few comments and suggested next steps :
- I would like to add an end date for the run since this is an important aspect of the new parallel run concept
- missing UI to navigate from one run to another.

PR #808, opened 2026-06-20
rfc: phase 4 — apache configs + infra scripts

mglst:
Phase 4 of the static-serving / multi-instance RFC (`docs/architecture/static-serving-and-deployment.md`). Builds on Phase 3 (#807, merged). Infra/ops only — no application Python changed.

## What's here

**`scripts/infra/`** (templates + per-server/per-instance setup)
- `apache-instance.conf` / `apache-main.conf` — vhost templates, `sed`-rendered (`@INSTANCE@`/`@PORT@`/`@DOMAIN@`)
- `energetica.service` — systemd template setting the full Phase-3 env contract
- `instance.json.tmpl` — initial per-instance config matching the `InstanceConfig` schema
- `setup-base.sh` / `setup-landing.sh` / `setup-instance.sh` — provisioning

**`scripts/`** (run from a workstation)
- `deploy-instance.sh` / `deploy-landing.sh` — rsync deploys
- `list-instances.sh` — systemd-based instance discovery

Removes the superseded `scripts/vps-setup.sh`. Updates the RFC status block and `DEPLOYMENT.md`.

## Key decisions

- **rsync, no git on the server.** `setup-instance.sh` provisions the box (dir, venv, `/etc/energetica/{slug}/instance.json`, vhost+TLS, *enabled-but-unstarted* unit); the **first `deploy-instance.sh`** ships the backend, installs deps, and starts the service. Resolves the RFC's `setup-instance.sh` step-2 TBD (clone vs symlink) in favour of neither.
- **Shared `energetica` user/group.** Instance services run as it and publish fragments into the `setgid` landing dir; `www-data` joins the group to read the `0o640` fragments (the readability fix that landed in #807).
- **Scoped sudoers drop-in** in `setup-base.sh` so the deploy user can run `sudo -u energetica pip …` and `sudo systemctl restart energetica-*` non-interactively.
- **SPA fallback scoped to `/app/*`** so a missing hashed asset returns a clean 404 instead of `index.html`-with-200 (which a browser would try to execute as JS).
- Instance discovery is via `systemctl list-unit-files 'energetica-*.service'`, never filesystem globbing (would wrongly include `energetica-landing`).

## Review

A two-angle review (bash correctness; Apache/systemd config) ran before pushing. Real findings fixed: `/app`-scoped fallback, sed-metachar escaping + JSON-breaking-char rejection for `--name`/`--starts-at`, slug validation in `deploy-instance.sh`, restart-failure diagnostics, idempotent venv creation, unit-enumeration filtering.

## Validation

`bash -n` clean on all six scripts. `shellcheck` / `apache2ctl` / `certbot` are unavailable in the dev sandbox, so **live conf syntax-checking and end-to-end validation (real subdomain, cookie isolation, fragment round-trip) are deferred to Phase 6**, as the RFC plans.

<h3>Greptile Summary</h3>

Phase 4 of the static-serving RFC: replaces the legacy monolithic `vps-setup.sh` with a layered `scripts/infra/` provisioning suite (setup-base → setup-landing → setup-instance) and matching workstation-side deploy scripts. All infra/ops-only; no application Python is modified.

- **Provisioning:** `setup-base.sh` installs packages and writes a tightly-scoped sudoers drop-in (per-instance pip path + `systemctl energetica-*`); `setup-instance.sh` pre-creates `.venv` owned by `energetica` so `sudo -u energetica pip install` can write into it, validates and rejects JSON-breaking chars in both `--name` and `--starts-at` before sed-rendering `instance.json`, and enables (but does not start) the systemd unit.
- **Apache configs:** `apache-instance.conf` serves the app bundle directly, scopes the SPA `FallbackResource` to `/app/*` only, and now includes a `/healthz` proxy rule so `deploy-instance.sh`'s health poll succeeds; `apache-main.conf` covers the pure-static apex landing with a 60-second TTL on manifest/fragment files.
- **Deploy scripts:** `deploy-instance.sh` guards the `jq` dependency upfront (preventing the previous silent 600-second timeout), omits `StrictHostKeyChecking=accept-new`, and polls `/healthz` after restart; `deploy-landing.sh` rsyncs the landing bundle while preserving instance-owned fragments.

Safe to merge — all previously flagged blocking issues (silent jq failure, missing /healthz proxy, venv ownership, NOPASSWD scope, --starts-at guard) are addressed in this revision.

The two issues from the prior review round that blocked deploys (/healthz not proxied and the jq silent-failure loop) are both fixed. Input validation, sudoers scoping, and venv ownership are all correct. The only new finding is a narrow timing edge case in the landing deploy rsync that cannot cause silent breakage during normal operation.

scripts/deploy-landing.sh — the missing `.well-known` exclude in the rsync is the one remaining item to address.

| Filename | Overview |
|----------|----------|
| scripts/deploy-landing.sh | Workstation-side landing deploy: rsyncs `dist-landing/` with `--delete`, excluding `instances.json` and `instances/`. Missing `--exclude='.well-known'`: certbot renewal challenge files placed in the landing root could be deleted by a concurrent deploy, potentially causing a renewal failure. |
| scripts/deploy-instance.sh | Workstation-side deploy: guards `jq` availability upfront (fixing the previous silent 600s false-negative), rsyncs backend + app bundle, pip-installs into the server venv, restarts the service, and polls `/healthz` (now proxied) for `status=ok`. No `StrictHostKeyChecking` override; slug and domain validated before use in remote paths. |
| scripts/infra/setup-instance.sh | Per-instance provisioner: creates deploy-owned app dir (2750), pre-creates `.venv` owned by energetica so the subsequent `sudo -u energetica python3 -m venv` succeeds, writes `instance.json` from template with JSON-breaking-char rejection for both `--name` and `--starts-at`, obtains TLS, and installs/enables the systemd unit. |
| scripts/infra/setup-base.sh | New per-server base provisioner: installs packages, creates shared `energetica` group/user, writes a scoped sudoers drop-in (per-instance pip path + `systemctl energetica-*` only), and sets up firewall and certbot hook. Idempotent, well-guarded. |
| scripts/infra/apache-instance.conf | Instance vhost template: serves static assets directly, WebSocket-upgraded Socket.IO via mod_rewrite, proxies `/api`, `/socket.io`, `/logout`, and `/healthz` to uvicorn. `/healthz` proxy rule (previously absent) is now present. SPA fallback scoped to `/app/*` via `<LocationMatch>`. |
| scripts/infra/setup-landing.sh | Provisions the apex landing DocumentRoot (setgid energetica, 2775) and its `instances/` fragment dir, obtains TLS via certbot webroot, then installs the full vhost from `apache-main.conf`. Domain validated against hostname regex before sed substitution. |

</details>

<h3>Flowchart</h3>

<a href="#gh-light-mode-only">

```mermaid
flowchart TD
    subgraph Server["Server (run as root)"]
        SB["setup-base.sh"]
        SL["setup-landing.sh"]
        SI["setup-instance.sh"]
        SB --> SL
        SL --> SI
    end
    subgraph Workstation["Workstation"]
        DI["deploy-instance.sh"]
        DL["deploy-landing.sh"]
    end
    SI -->|first deploy starts service| DI
    SL -->|ships landing bundle| DL
    subgraph Apache["Apache (www-data)"]
        AL["apex vhost"]
        AI["instance vhost"]
    end
    DL -->|rsync| AL
    DI -->|rsync| AI
```

</a>
<a href="#gh-dark-mode-only">

```mermaid
%%{init: {'theme': 'base', 'themeVariables': {"darkMode": true, "background": "#0d1117", "primaryColor": "#21262d", "primaryTextColor": "#e6edf3", "primaryBorderColor": "#8b949e", "lineColor": "#8b949e", "textColor": "#e6edf3", "edgeLabelBackground": "#161b22", "actorBkg": "#21262d", "actorBorder": "#8b949e", "actorTextColor": "#e6edf3", "actorLineColor": "#8b949e", "signalColor": "#8b949e", "signalTextColor": "#e6edf3", "noteBkgColor": "#373320", "noteBorderColor": "#d4a72c", "noteTextColor": "#f0e6c0", "labelBoxBkgColor": "#21262d", "labelBoxBorderColor": "#8b949e", "labelTextColor": "#e6edf3", "loopTextColor": "#e6edf3", "activationBkgColor": "#30363d", "activationBorderColor": "#8b949e"}}}%%
flowchart TD
    subgraph Server["Server (run as root)"]
        SB["setup-base.sh"]
        SL["setup-landing.sh"]
        SI["setup-instance.sh"]
        SB --> SL
        SL --> SI
    end
    subgraph Workstation["Workstation"]
        DI["deploy-instance.sh"]
        DL["deploy-landing.sh"]
    end
    SI -->|first deploy starts service| DI
    SL -->|ships landing bundle| DL
    subgraph Apache["Apache (www-data)"]
        AL["apex vhost"]
        AI["instance vhost"]
    end
    DL -->|rsync| AL
    DI -->|rsync| AI
```

</a>

<details><summary><h3>Comments Outside Diff (2)</h3></summary>

1. `scripts/infra/apache-instance.conf`, line 383-387 ([link](https://github.com/felixvonsamson/energetica/blob/51ef8d569f5b6970fe8a256d8813922399d3df7e/scripts/infra/apache-instance.conf#L383-L387)) 

   <a href="#">P1</a> **`/healthz` is not proxied — every deploy times out for 600 s then fails**

   `deploy-instance.sh` polls the `/healthz` endpoint (line 140) after restart. That route is registered at the top-level prefix `""` in `energetica/routers/health.py:79`, so it lives at `/healthz` on the uvicorn port — not under `/api`. The only proxy rules here forward `/socket.io/`, `/api`, and `/logout`; `/healthz` receives no matching rule, Apache finds no static-file match and no `/app/…` fallback, and returns 404. `curl -f` suppresses the body on 4xx, `HZ` stays empty every iteration, and the loop exhausts all 600 s before exiting with "did not reach status=ok within 600s".

   Adding a `ProxyPass /healthz` + `ProxyPassReverse /healthz` pair alongside the existing `/logout` rules would route the health-check to uvicorn.

2. `scripts/deploy-instance.sh`, line 220-233 ([link](https://github.com/felixvonsamson/energetica/blob/51ef8d569f5b6970fe8a256d8813922399d3df7e/scripts/deploy-instance.sh#L220-L233)) 

   <a href="#">P1</a> **Silent `jq` dependency causes 600-second false-negative deploy**

   `jq` is called on the **local** machine (where `deploy-instance.sh` runs), not the remote server. When `jq` is absent, `jq -r '.status'` fails, the `2>/dev/null || echo ""` silently swallows the error, `STATUS` stays `""`, and `SCHED_ERRS` stays `0`. Neither branch exits early — the loop exhausts all 600 seconds then prints `/healthz did not reach status=ok within 600s`, making a successful deploy look like a failure. The operator has no indication that the service is actually running. `setup-base.sh` installs `jq` on the server but there is no equivalent check or documentation requirement for the workstation.
</details>

<sub>Reviews (5): Last reviewed commit: ["review: reject JSON-breaking chars in --..."](https://github.com/felixvonsamson/energetica/commit/e5c0d626a94983009139bb20fa4648c947a9ecab) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=38818761)</sub>


mglst:
Addressed in `51ef8d56`:

- **P1 — venv permission mismatch** (`setup-instance.sh`): fixed. `.venv` is now pre-created owned by `energetica` (mode `0750`) before `python3 -m venv` populates it, so the service user can both create it and write into it during `deploy-instance.sh`'s `sudo -u energetica pip install`. The code dir `$APP_DIR` stays deploy-owned (least privilege) — I went this route rather than create-as-root-then-chown to avoid a recursive chown.
- **`(energetica) NOPASSWD: ALL`** (`setup-base.sh`): scoped to `/var/www/energetica-*/.venv/bin/pip *`. The deploy user can no longer use the service account for anything but pip.
- **`systemctl is-active *`**: I took a slightly different path than the suggestion. `is-active` is a read-only query that needs no root, so I dropped `sudo` from it in `deploy-instance.sh` and removed it from sudoers entirely. (Heads-up: the verbatim suggestion `is-active energetica-*` would have *broken* the call — the script passes `--quiet energetica-<slug>`, so the `--quiet` arg wouldn't have matched the pattern.) `status`/`journalctl` narrowed to `energetica-*`.
- **`StrictHostKeyChecking=accept-new`** (both deploy scripts): removed the override so SSH's default surfaces an unexpected first-seen key; a host the operator already configured connects silently and CI pre-populates `known_hosts`.
- **Step-8 numbering**: kept the RFC-aligned 1–10 numbering and made the step-8 no-op (certbot renewal hook, installed once by `setup-base.sh`) explicit so the sequence no longer appears to skip.

@greptileai

mglst:
@greptileai

felixvonsamson (approved):
Optional : Could be usefull to add a description to a run to potentially show specific information on the frontend or at least to have the context of the run in the backend when managing all the runs.
I suggest to not have a strict naming convention for the names but to have tthe description add some information about the specificity of the run (e.g. "extreme climate", "coal paradise"...)

mglst:
Round-2 fixes in `d6215db6`:

- **P1 — `/healthz` not proxied** (`apache-instance.conf`): confirmed `health_router` is mounted at the uvicorn root (`energetica/routers/__init__.py:202` — no `/api/v1` prefix, unlike the API routers), so Apache had no rule for it and the deploy health poll always 404'd. Added `ProxyPass`/`ProxyPassReverse /healthz` alongside `/logout`.
- **P1 — silent local `jq` dependency** (`deploy-instance.sh`): added a `command -v jq` preflight that fails fast with an install hint, so a missing `jq` can no longer turn a successful deploy into a 600s false-negative.
- **`$DOMAIN` not sed-escaped** (`setup-instance.sh`, `setup-landing.sh`): a DNS hostname can't actually contain a sed metacharacter, but I added a hostname-format check at the door — it closes the concern and also rejects typos that would otherwise only surface later as a certbot/Apache failure.

Good catch on the two health-check issues compounding — together they'd have made every first deploy look broken. Both verified: `/healthz` route location confirmed in code, domain regex tested against the metachar cases.

mglst:
Fixed in `e5c0d626`. Correct catch — the guard was on `--name` only, and `sed_escape` handles sed metachars (`&`/`/`/`\`) but not a JSON-breaking `"`, so a `--starts-at` with a quote would have rendered invalid JSON → fail-closed 403. Added the matching `case "$STARTS_AT"` guard, so the implementation now matches what the description claimed.

PR #820, opened 2026-06-29
fix: reclaim local & against-prod dev workflows broken by the multi-instance refactor

mglst:
Two dev workflows broke when the static-serving / server-wide-accounts refactor (RFC Phases 2–5) cut production over to the landing-apex + instance-subdomain topology. Both surfaced as `Unexpected token '<', "<!DOCTYPE"... is not valid JSON`. This PR reclaims both. Scope is strictly **dev/local workflows** — no production behaviour changes.

## Part 1 — local full-stack loop (`python main.py --env dev` + `bun dev`)

**Accounts store had no dev fallback.** `accounts/db.py` defaulted to `/var/lib/energetica/accounts.db`; `_connect()` `mkdir`s the parent, which `PermissionError`s under root-owned `/private/var` on macOS, so the admin bootstrap, signup, and login all failed locally.
→ Default to `instance/accounts.db` (repo-local, co-located with the pickle/secret key, wiped together by `--rm_instance`). Production sets `ENERGETICA_ACCOUNTS_DB_PATH` explicitly via systemd, so the default is dev-only. `migrate-to-server-accounts.py` pins the prod path explicitly so it never inherits the dev default.

**Vestigial redirect produced the `<!DOCTYPE` error.** The `log_action` middleware turned any unauthenticated GET into `303 → /app/login`; Vite doesn't proxy `/app`, so a `fetch` follows it onto the SPA `index.html`. Post-Phase-5 FastAPI serves no HTML, so the redirect is dead weight in prod too.
→ Delete the redirect; the API returns `401` JSON and the SPA (`routes/__root.tsx`, via `/auth/me`) owns the client-side redirect. Regression test added (red-first: `303 == 401`).

## Part 2 — against-prod frontend modes (`bun run dev:game` / `dev:edu` / `dev:ethz`)

These run local Vite proxied at a live backend. Each `.env.{mode}` hard-coded `VITE_BACKEND_URL` at an **apex** — but the cutover made every apex a pure-static landing whose `FallbackResource → index.html` answers all paths, so proxied `/api` came back as HTML. The game backend now lives on a **rotating instance subdomain** (`{slug}.{apex}`), so baking a slug would rot each season.
→ `.env.{mode}` now sets only the **stable** `VITE_APEX_DOMAIN`; `vite.config.ts` discovers the current slug at dev-server startup from the apex's public `instances.json` (newest advertised — the ordering the landing picker uses). No slug is hardcoded; it self-updates across seasons. Explicit `VITE_BACKEND_URL` still wins; default stays `localhost:8000`; runs only for `serve` (never at build); fails loud on an empty/unreachable manifest.

## Scope & non-goals

- Restores the **single-instance localhost loop** and the **against-prod modes** only. A local harness that *mimics* the apex+instance topology (subdomains-on-localhost behind a reverse proxy) was considered and **deferred** — its unique value (cross-origin landing↔instance↔lobby flows, the parent-domain SSO cookie) is mostly not built yet; it should land with the lobby work.
- No ADR — the operative principle (*code defaults target dev; prod is explicit via systemd*) lives as a code comment in `db.py`.

## Out of scope (flagged, not fixed here)

`package.json`'s `deploy:game` / `deploy:edu` / etc. still call `bash scripts/deploy.sh`, which was **deleted** in Phase 5 (replaced by `deploy-instance.sh` / `deploy-landing.sh`). Those *deployment* aliases are broken too, but that's a separate concern from dev workflows.

## Verification

- Backend: 30 affected tests pass (accounts, login, signup, change-password, templates); `ruff` clean. Smoke-tested the real dev default with `ENERGETICA_ACCOUNTS_DB_PATH` unset → resolves to `instance/accounts.db`, no `PermissionError`.
- Frontend: `bun run dev:game` resolves `energetica-game.org → mar-27-2026` from the live manifest; plain `bun dev` stays on localhost with no fetch; `tsc` clean.

<h3>Greptile Summary</h3>

This PR restores two broken dev workflows after the multi-instance refactor: the local full-stack loop (`python main.py --env dev` + `bun dev`) and the against-prod frontend modes (`bun run dev:game` / `dev:edu` / `dev:ethz`). Both failures manifested as `Unexpected token '<', "<!DOCTYPE"...` JSON parse errors.

- **Accounts DB default** changed from `/var/lib/energetica/accounts.db` to the CWD-relative `instance/accounts.db`, eliminating the `PermissionError` on macOS; production always overrides via `ENERGETICA_ACCOUNTS_DB_PATH` in systemd. The migration script and test fixtures are updated to stay pinned to the prod path and a per-test tmp file, respectively.
- **Vestigial 303 redirect** removed from the `log_action` middleware — the API now only speaks JSON on 401; the SPA owns the client-side redirect. A regression test (red-first) locks in the JSON 401 contract.
- **`.env.{mode}` files** replaced hardcoded apex URLs with `VITE_APEX_DOMAIN`; `vite.config.ts` now discovers the current season's instance slug at dev-server startup by fetching `instances.json` from the apex, with `VITE_BACKEND_URL` as an explicit override and `localhost:8000` as the plain-dev fallback.

Safe to merge — all changes are scoped to dev workflows; production behaviour is unchanged.

The accounts DB default only affects local dev (production always supplies the path via systemd); the 303 redirect removal is verified by a new regression test; the Vite config apex-discovery logic is guarded behind command === serve so it never runs during builds; and the migration script is correctly hardened to never inherit the new dev-friendly default. No production code paths are touched.

| Filename | Overview |
|----------|----------|
| frontend/vite.config.ts | Major additions: async resolveBackendUrl (fetches instances.json to discover live slug), devRootRedirect plugin mirroring Apache root redirect, rewriteSetCookieForLocalhost cookie scrubber for /api and /logout. Config function made async. |
| energetica/accounts/db.py | Default DB path changed from /var/lib/energetica/accounts.db to instance/accounts.db (CWD-relative). Safe for dev; prod always overrides via ENERGETICA_ACCOUNTS_DB_PATH. |
| energetica/routers/__init__.py | Removed 303 redirect for unauthenticated GET requests; also removes the now-unused RedirectResponse import. Clean removal of dead code. |
| scripts/migrate-to-server-accounts.py | Hardcodes the prod accounts-db default in argparse and unconditionally sets the env var, preventing accidental inheritance of the new dev-oriented library default. |
| tests/conftest.py | Docstring updated to reflect new dev default (instance/accounts.db); autouse fixture correctly redirects all tests to a per-test tmp path. |
| tests/integration/test_templates_routes.py | Added regression test verifying unauthenticated protected GETs return JSON 401, not a 303 redirect. |

<sub>Reviews (3): Last reviewed commit: ["fix(dev): dev-only / → /app/ redirect; s..."](https://github.com/felixvonsamson/energetica/commit/0c6962cc2480759bbc632b9445f737477920e4fe) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=40627712)</sub>

### DIFF
--- a/docs/architecture/authentication.md
+++ b/docs/architecture/authentication.md
@@ -55,7 +55,7 @@ To authenticate via API:
 
 **Middleware:** `energetica/routers/__init__.py`
 
--   Unauthenticated GET requests that return 401 are redirected to `/app/login`
+-   The API only ever speaks JSON — it never issues browser redirects. An unauthenticated request returns a `401` (JSON) from the endpoint; the SPA owns the redirect to `/app/login` client-side (`frontend/src/routes/__root.tsx`, driven by `/auth/me`). After the Phase 5 cutover FastAPI serves no HTML, so a server-side redirect could only land a `fetch` client on Apache's (or, in dev, Vite's) `index.html` — the `Unexpected token '<', "<!DOCTYPE"...` break.
 -   POST/PUT/PATCH/DELETE requests require authentication (enforced per-endpoint)
 
 **Authentication enforcement:** Endpoints use FastAPI `Depends()` to require authentication:
--- a/docs/getting-started/installation.md
+++ b/docs/getting-started/installation.md
@@ -24,16 +24,22 @@ source .venv/bin/activate && python main.py --env dev
 
 ### Run the Frontend
 
-Use existing backend:
+Against the live production game backend (auto-discovers the current instance — see below):
 
 ```bash
-VITE_BACKEND_URL=https://energetica-game.org bun dev
+bun run dev:game
 ```
 
-Use http://localhost:8000 as backend:
+Against a local backend (`http://localhost:8000`):
 
 ```bash
 bun dev
+```
+
+Against one specific backend, overriding everything:
+
+```bash
+VITE_BACKEND_URL=https://mar-27-2026.energetica-game.org bun dev
 ```
 
 Or configure `/frontend/.env.example`.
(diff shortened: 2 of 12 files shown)


felixvonsamson (approved):
Did not read but I trust you. lmk if it's something I should look into

Issue #817, opened 2026-06-28
Lobby Phase C — cutover flag day (SSO + retire instance auth + switcher)


mglst:
Part of #814. Design: `docs/architecture/lobby.md` § Phasing. **Coordinated flag-day deploy** (like the Phase-5 cutover), not a quiet merge.

- [ ] Flip instances to the shared secret + parent-domain cookie.
- [ ] Entry gate: move local-`User` auto-provision out of `/login` into the authenticated-entry path, access-policy-gated; clean 403 on denial.
- [ ] Retire instance login/signup pages and credential endpoints; unauth → redirect to `lobby.{apex}/login?return={slug}`; repoint the instance-local `/sign-up` redirect at the lobby.
- [ ] In-run switcher (top-right, next to Logout) — your settled runs + "Open lobby"; fed by the instance's own `my-runs` (#814 Phase A). **This is the rescoped #810.**
- [ ] Landing: CTAs → lobby; remove `AdvertisedRuns` / `useAdvertisedInstances`; retire `primaryInstanceAppHref` / `instanceSignupHref`.
- [ ] Global logout (clears the one `.{apex}` cookie).
- [ ] Land the #813 CI guard (forbid raw-HTML rendering in UGC surfaces).
- [ ] **One-time forced global re-login** — document in `DEPLOYMENT.md`; coordinate.

mglst:
Completed and shipped — merged in #841 (`Lobby Phase C cutover: SSO entry gate, retire instance auth, in-run switcher`).

Delivered:
- **C1** — instance flip: SSO cookie renamed `session` → `energetica_session` (resolves the #816 B→C dual-cookie wrinkle), shared-secret cookie carrying immutable `account_id`, and the `GET /auth/me` **entry gate** (validates cookie, enforces per-instance access policy fail-closed, auto-provisions local `User` under `engine.lock`).
- **C2** — retired instance `/auth/{login,signup,logout,change-password}` + legacy root `GET /logout`; Socket.IO reads the renamed cookie.
- **C3** — frontend: unauth → cross-origin redirect to `lobby.{apex}/login?return={slug}`; global logout; in-run switcher; DNS-label slug validation; landing CTAs → lobby; `AdvertisedRuns` removed.
- Schema generator merges lobby routes into `api.generated.ts`; DEPLOYMENT.md cutover runbook (one-time forced global re-login).

Deferred carve-outs, tracked separately:
[...]

Issue #816, opened 2026-06-28
Lobby Phase B — stand up the lobby service (backend + bundle + infra)


mglst:
Part of #814. Design: `docs/architecture/lobby.md` § Architecture / Frontend / Infrastructure.

Develops alongside untouched instances; nothing in the instances changes yet.

- [ ] Lobby backend (uvicorn): `signup` (account-only, ADR-0003), `login` (sets `domain=.{apex}` cookie, shared secret), `change-password`, `logout`, `my-runs` (reads `accounts.db` + fragments).
- [ ] **Server-wide signup gate:** new `/etc/energetica/server.json` with `signups_enabled`, read fresh by the lobby (server-wide analog of the per-instance `disable_signups`, which the lobby can't reuse). `setup-base.sh` writes the initial file.
- [ ] **Closed-enrollment bootstrap:** admin path to seed accounts directly into `accounts.db` (`accounts.get_or_create_account_id`), replacing per-instance `players.txt` for account creation.
- [ ] Lobby frontend bundle (sibling to landing): `/login`, `/signup`, picker with all states (logged-out ±runs, logged-in with/without runs).
- [ ] Validate `?return=` against the live instance list (open-redirect prevention).
- [ ] Infra: `infra/setup-lobby.sh`, `infra/apache-lobby.conf`, `infra/energetica-lobby.service`, `deploy-lobby.sh`; `setup-base.sh` creates the shared secret + `server.json`; DNS for `lobby.{apex}`.

Note: `my-runs` is the *same read* served by both the lobby (here) and each instance (Phase A / #815), each on its own origin — no CORS.

mglst:
### Design grill outcomes (pre-implementation)

Resolved before coding; scope trims and key decisions:

- **Module boundary:** lobby is a top-level `lobby/` package (`create_lobby_app`, entry `main_lobby.py`) reusing `energetica.accounts` / `instance_config`. To keep the import engine-free, `energetica/__init__.py` becomes lazy (engine construction + heavy router/socketio imports move into `create_app`) and the signing/cookie/password primitives are extracted into a leaf `energetica/utils/session.py` (no game-model imports); `utils/auth.py` re-exports them and keeps the `User`/`Player` deps. A full identity-package extraction is a **follow-up** (too large for B).
- **Session payload = `account_id`** (not `username`) so future username changes cost nothing. Lobby mints it in B; instance side flips at C's forced re-login. ADR-0002 amended.
[...]

PR #836, opened 2026-07-05
Lobby Phase B (backend): engine-free identity layer + lobby service

mglst:
Part of #816 (lobby Phase B). **This PR is the backend slice**; the frontend bundle and infra scripts follow in separate PRs on the same issue. Develops alongside untouched instances — nothing in the instances changes yet.

## What's here

**Engine-free identity layer (refactor)**
- `energetica/utils/session.py` — new game-model-free leaf: cookie-signing secret, itsdangerous serializer, password hashing, session-cookie helpers. `utils/auth.py` re-exports these (game-side call sites unchanged) and keeps the `User`/`Player`-coupled request deps. `accounts/db.py` imports the leaf, severing `accounts → auth → player`.
- `energetica/__init__.py` — the heavy game graph (routers, socketio, tick loop, domain models) moves from module scope into `create_app`, so importing the identity layer / the lobby loads none of it. A dormant, light `GameEngine` object is still built at import (the ORM binds it at model-definition time); full severance is a follow-up identity-package extraction. Guarded by `tests/unit/test_module_boundary.py`.

**Shared reads + config**
- `accounts.get_account_by_id`; `energetica/my_runs.py` (the my-runs join, now shared by the instance route *and* the lobby); `instance_config.list_advertised_fragments`.
- `energetica/server_config.py` — `server.json` signup toggle, read fresh, **fail-closed** (`StrictBool`; missing/malformed → signups disabled) per ADR-0003.

**Lobby service**
- `lobby/` — `create_lobby_app` mounts `/api/v1/auth/{login,signup,change-password,logout}` and `/api/v1/lobby/my-runs`, **reusing the game's request schemas** so `api.generated.ts` covers the lobby with no new type generation. Signup is account-only (ADR-0003), gated by `server.json`, auto-logs-in. Session cookie signs the **`account_id`** scoped to `.{apex}` (ADR-0002); identity resolves straight against `accounts.db` (no game `User`). `main_lobby.py` uvicorn entry.

## Decisions (details on #816)
- **Module boundary is pragmatic**, not absolute — dormant engine object stays; the domain/routers/socketio/tick loop don't load. Full severance = follow-up C extraction.
- **`?return=` is handled client-side** in the SPA (login is an XHR returning JSON, like the game) → schemas identical, types reused. No server open-redirect: the URL is always *constructed* to the fixed apex, never a passthrough.
- ADR-0002 amended: session payload = `account_id` (instances adopt it at the Phase C forced re-login).

## Testing
- ~30 new tests (module-boundary guard, shared reads, `server.json` fail-closed, full lobby flow via `TestClient`). Whole-repo `pyright` clean; `ruff` clean on all new files; `generate-types` produces **no drift**.
- Full suite green except 11 pre-existing failures unrelated to this work (external-URL `daily_quiz` checks, a bare-`python` subprocess in `test_server_runs`).

## Not in this PR
Frontend lobby bundle (#816), infra (`setup-lobby.sh` / `apache-lobby.conf` / `energetica-lobby.service` / `deploy-lobby.sh`, `setup-base.sh` shared-secret + `server.json`), and the deploy precondition guard. CI XSS guard and the instance cutover remain Phase C (#817).

<h3>Greptile Summary</h3>

This PR adds the backend lobby service and splits shared identity/session code out of the game app. The main changes are:

- Engine-free session and password helpers for shared account auth.
- Lobby auth routes for login, signup, password change, logout, and my-runs.
- Shared account and run lookup helpers used by both lobby and instance routes.
- Server-wide signup configuration with fail-closed parsing.
- Lazy game imports so lobby/account modules avoid loading the heavy game graph.

This looks safe to merge.

- No blocking issues found in the changed code.

| Filename | Overview |
|----------|----------|
| lobby/session.py | Adds lobby helpers to set, clear, and decode the account-id session cookie. |
| energetica/my_runs.py | Adds the shared my-runs resolver and timestamp parsing for stored memberships. |
| energetica/utils/session.py | Moves cookie signing, password hashing, and session-cookie helpers into a shared leaf module. |
| lobby/routers.py | Adds lobby credential routes and the authenticated my-runs endpoint. |

<sub>Reviews (2): Last reviewed commit: ["fix: make my-runs tolerant of a malforme..."](https://github.com/felixvonsamson/energetica/commit/10ecfd5d6177d607dd42e09dd6b13aefeee218d5) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=41931691)</sub>


mglst:
Thanks — addressed both:

**P2 (my-runs 500 on bad timestamp) — fixed** in 10ecfd5d. `record_membership` normalises every write to aware UTC and is the sole writer, so this isn't reachable through current code, but the read already tolerates *other* bad data (it skips a run whose fragment is gone), so it was inconsistent for one malformed row to 500 the whole response and hide every run. `resolve_my_runs` now recovers a naive timestamp as UTC and skips a truly unparseable one (logged). Added two tests inserting raw rows that bypass the write-side normalisation.

**P1 (parent cookie survives logout) — no change needed; it's by design.** Logout is centralised at the lobby: `clear_lobby_session_cookie` deletes the `.apex` cookie with the matching `Domain`/`path` (covered by `test_logout_clears_session`). Per-run logout of a shared cookie is explicitly *not offered* (design § Flows "Logout (global)"). And there is no phase where a leftover `.apex` cookie authenticates on an instance after logout:
- **Phase B (this PR):** instances run their pre-cutover code, signing/validating a *host-only* cookie with their *per-instance* secret. They can't validate the lobby's shared-secret `.apex` cookie at all, so it never authenticates there; their own logout correctly clears their own host-only cookie.
- **Phase C:** instances adopt the shared cookie and their login/logout is retired (unauth → redirect to lobby), so the only logout is the lobby's domain-aware clear.

No instance code changes here (instances are untouched in Phase B and don't yet know the apex). I've noted "instance logout is retired at C; the lobby's domain-aware clear is the only logout" as a Phase C invariant.

felixvonsamson (approved):

PR #837, opened 2026-07-05
Lobby Phase B (frontend): login / signup / picker bundle

mglst:
Part of #816 (lobby Phase B). Stacks on #836 (lobby backend) — merge that first; if it lands squashed, I'll rebase.

## What this adds

The lobby SPA bundle, a third Vite bundle mirroring the landing-bundle pattern (`index.lobby.html` → `src/main-lobby.tsx`, `src/routes-lobby/`, `routeTree.lobby.gen.ts`, `dist-lobby/`, `tsconfig.lobby.json`, `dev:lobby` / `build:lobby` scripts, typecheck + eslint wiring).

- **Picker (`/`)** — logged-out: brand hero + "runs open to join" (each card routes through `/login?return={slug}`); logged-in: "your runs" from `GET /api/v1/lobby/my-runs` (icon-tile cards, joined date, Continue) + "open runs" from the same-origin `instances.json` manifest filtered against memberships; logged-in-no-runs empty state.
- **`/login` / `/signup`** — mirror the game's forms against the lobby backend (signup is account-only, ADR-0003; `SIGNUP_DISABLED` surfaces via the shared error mapping).
- **`?return={slug}` resolved client-side** (per the #816 design grill): validated against the runs the SPA knows (my-runs ∪ manifest) and only ever interpolated into the fixed `https://{slug}.{apex}/app` template, apex derived from `window.location.hostname` — never a passthrough, no open redirect. Unknown slugs fall through to the picker.
- **Auth probe**: the lobby has no `/auth/me`; `my-runs` doubles as the probe (401 → logged out, other errors → retryable error state).
- Light providers (Theme + a lobby-local QueryClient); reuses the game's `api-client`, generated types (`/api/v1/lobby/my-runs` was already in `api.generated.ts` — no type drift) and error envelopes.
- Dev server: proxies `/api` → `main_lobby.py` (:8001) and serves `index.lobby.html` for SPA paths (Vite otherwise answers with the game's `index.html` and chokes on its MDX imports); `/api` excluded from the fallback so browser navigations still reach the backend.

## Verified

[...]

PR #838, opened 2026-07-05
Lobby Phase B (infra): vhost, unit, setup + deploy scripts

mglst:
Part of #816 (lobby Phase B). Stacks on #837 (frontend bundle), which stacks on #836 (backend).

## What this adds

`scripts/infra/` gains the lobby siblings of the landing/instance files, plus `scripts/deploy-lobby.sh`:

- **`apache-lobby.conf`** — serves `dist-lobby` (SPA `FallbackResource`, disabled under `/assets/` so pruned hashed assets 404 instead of being cached as HTML), proxies `/api` to the lobby uvicorn, and **`Alias`es the live `instances.json` from the shared landing dir** (same-origin manifest for the picker — no CORS, no second copy). `X-Forwarded-Proto` set so the parent-domain cookie gets `Secure`.
- **`energetica-lobby.service`** — `ENERGETICA_APEX_DOMAIN` (cookie scope, ADR-0002) + explicit shared-state paths (`secret_key.txt`, `accounts.db`, landing dir, `server.json`); runs `main_lobby.py` as the shared service user.
- **`setup-lobby.sh`** — dir + venv + TLS + vhost + unit (enabled, not started), preconditions on the setup-base/landing artifacts, **port-uniqueness check** (a collision with an instance is insidious: the lobby unit crash-loops while Apache proxies lobby `/api` to the *game* backend, which also answers `my-runs` with 401 — even the health check would pass; the old DEPLOYMENT.md example provisioned an instance on the lobby's default 8001). Re-runs no longer clobber the live HTTPS vhost with the temp ACME vhost.
- **`deploy-lobby.sh`** — **hard-refuses (fail-closed, before anything ships) unless `instance_membership` exists in `accounts.db`** — the #816-pinned Phase A precondition (otherwise every existing player silently sees zero runs). Then rsync backend + bundle, pip install, restart, health check (api 401 unauthenticated + SPA 200). No `VITE_APEX_DOMAIN`: the lobby derives its apex from its hostname at runtime.
- **`setup-base.sh`** — also provisions the server-wide signing secret (never rotated on re-run) and `/etc/energetica/server.json` (`{"signups_enabled": true}`); idempotent, so existing boxes just re-run it.
- **`lobby` is now a reserved slug** in `setup-instance.sh` / `deploy-instance.sh` (would collide with the lobby vhost/dir).
- DEPLOYMENT.md updated (lobby setup/deploy sections, one-port-per-service rule).

## Reviewer call-outs (deliberate tradeoffs)

- **Sudoers grant for the precondition check**: the deploy user can't read `accounts.db`, so `setup-lobby.sh` grants `python3 -c *` **as the `energetica` service user only** (never root). This adds no capability beyond the existing pip rule (pip installs already execute arbitrary code as that user), but it *is* a broad-looking line — the alternative is a `/healthz`-style preflight in the lobby backend, which #816 pinned against in favor of refusing at deploy time. Happy to switch if you prefer.
- The deploy health check keys on "`/api/v1/lobby/my-runs` answers 401 unauthenticated" since the lobby app has no `/healthz`; a tiny `/healthz` route would be a nicer contract but touches the #836 backend, so it's left as a possible follow-up.

Everything mirrors the existing landing/instance script patterns (colors, flags, rsync exclusions, sudoers style); `bash -n` clean, template rendering + port-conflict grep + the sqlite precondition snippet tested against fixtures (present / absent table / missing db).

<h3>Greptile Summary</h3>

This PR adds the lobby deployment and provisioning path. The main changes are:

- New lobby Apache vhost, systemd unit, setup script, and deploy script.
- Shared signing secret and server config provisioning in base setup.
- Lobby slug reservation for instance setup and deploy.
- Deployment docs for lobby setup, deploy, and port separation.

This looks safe to merge.

- No blocking issues found in the changed code.

| Filename | Overview |
|----------|----------|
| scripts/deploy-lobby.sh | Adds the lobby deploy flow with Phase A precondition checks, bundle sync, restart, and live health probes. |
| scripts/infra/setup-lobby.sh | Adds lobby provisioning for directories, venv, sudoers, TLS, Apache, and systemd. |
| scripts/infra/apache-lobby.conf | Adds the lobby vhost with SPA serving, API proxying, asset caching, and same-origin manifest aliasing. |
| scripts/infra/energetica-lobby.service | Adds the lobby service unit with explicit shared-state paths and apex-domain configuration. |
| scripts/infra/setup-base.sh | Adds idempotent provisioning for the shared session secret and server-wide config. |
| scripts/deploy-instance.sh | Updates instance deploy validation to reserve the lobby slug. |
| scripts/infra/setup-instance.sh | Updates instance setup validation to reserve the lobby slug and enforce DNS-label length. |
| scripts/DEPLOYMENT.md | Documents the lobby setup and deploy workflow. |

<sub>Reviews (3): Last reviewed commit: ["fix: manifest health probe + 63-char slu..."](https://github.com/felixvonsamson/energetica/commit/b48d8e8ee2016fdf627781536b60a2094982f024) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=41960437)</sub>

### DIFF
--- a/scripts/DEPLOYMENT.md
+++ b/scripts/DEPLOYMENT.md
@@ -15,9 +15,11 @@ On the VPS, as root:
 
 ```bash
 sudo bash scripts/infra/setup-base.sh --deploy-user deploy   # Apache, Python, certbot, firewall,
-                                                             # `energetica` group/user, shared dirs
+                                                             # `energetica` group/user, shared dirs,
+                                                             # shared secret + server.json
 sudo bash scripts/infra/setup-landing.sh --domain energetica-game.org   # apex vhost + TLS
-sudo bash scripts/infra/setup-instance.sh autumn-2025 8001 --domain energetica-game.org  # vhost+TLS+unit
+sudo bash scripts/infra/setup-lobby.sh --domain energetica-game.org     # lobby vhost+TLS+unit
+sudo bash scripts/infra/setup-instance.sh autumn-2025 8002 --domain energetica-game.org  # vhost+TLS+unit
 ```
 
 `setup-instance.sh` provisions the box (directory, venv, `/etc/energetica/{slug}/instance.json`,
@@ -35,6 +37,7 @@ From your local machine:
 ```bash
 ./scripts/deploy-instance.sh --server energetica-game --instance autumn-2025 --domain energetica-game.org
 ./scripts/deploy-landing.sh  --server energetica-game --domain energetica-game.org
+./scripts/deploy-lobby.sh    --server energetica-game --domain energetica-game.org
 ./scripts/list-instances.sh  --server energetica-game
 ```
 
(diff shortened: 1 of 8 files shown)


mglst:
Rebased onto the updated #837. All three Greptile P1s addressed in 90b8751b:

- **Sudo path mismatch** — the precondition check now invokes `/usr/bin/python3` fully qualified, matching the sudoers grant exactly.
- **Health check skips shared state** — added a third probe: login with a nonexistent user must answer `401 USER_NOT_FOUND`, which requires an actual `accounts.db` read (my-runs rejects at the cookie check before touching shared state; a broken db path answers 500 to this probe instead). The probe username is 22 chars — above signup's 18-char cap — so it can never collide with a real account. Verified against the real lobby backend locally.
- **Bundle permissions** — the `dist-lobby` rsync ships `--chmod=D755,F644` so a restrictive local umask can't deploy files www-data can't read.

mglst:
Manifest probe added in b48d8e8e: the health check now also requires `https://lobby.{apex}/instances.json` → 200 before reporting healthy, with a targeted operator hint on failure (missing → restart an instance to republish; 403 → check file permissions in the landing dir). Since the Phase A precondition already guarantees an instance deployment happened, a non-200 here is a genuine Alias/permissions fault rather than a fresh-box false alarm.

Same commit also caps the `setup-instance.sh` / `deploy-instance.sh` slug checks at the 63-char DNS label limit, matching the frontend's `isValidRunSlug` from #837.

felixvonsamson (approved):

Issue #979, opened 2026-08-09
Land the docs-reference checks in CI


mglst:
## Question

Build and land the three docs-integrity checks designed in #943, as one CI job.
Nothing to decide — the rule sets are ratified there. `docs-policy-review/proto-minimal.py`
on `research/docs-policy-review` is the working prototype of check 1.

**1. `docs:refs`.** A backticked token in `docs/**.md` is a repo-path claim iff it
matches `^[\w./@-]+$`, contains a `/`, and does not start with `/` or `@`. It passes
if it names a tracked file or directory, or sits under `instance/`, `checkpoints/`,
`frontend/dist`, `energetica/static/` or `.venv/`. Relative markdown links must
resolve; root-absolute `/…` links are errors; anchors in `.md` targets are checked.
Bare basenames are out of scope by decision. `docs/agents/*` is not scanned.

**2. Inverted citations.** Every `docs/**.md` path cited from code resolves, including
its `#fragment` anchor. Scans the repo outside `docs/`.

**3. Orphans.** Every doc is linked from `docs/README.md`, linked from `CONTRIBUTING.md`,
cited from code, or is an ADR. Direct reachability, not transitive. `docs/agents/*` excluded.

Blocked by #946, which makes the tree green: measured today the three checks fail 20
times in surviving text, and #946 owns every one of those repairs. **Land with no
allowlist and no suppression** — if the job is not green on arrival, the repairs are
incomplete, not the rules.

Follow `scripts/guard-no-raw-html.ts` for shape: a single script, a module docstring
that states what the rule protects and what it deliberately does not cover, and a
failure message that tells the reader how to fix rather than how to silence.

Part of #901.

mglst:
**Two ADR checks belong in this ticket, decided in #944.**

1. **No two ADR files share a number.** `docs/adr/` numbering raced: three parallel branches scanned a `main` whose highest was `0003`, and two independently allocated `0004`/`0005`. The convention ("scan for the highest and increment") is race-prone by construction, and #944 deliberately declined to write prose about it — per #904, enforcement belongs to tooling. This check is the whole mechanism. It works at PR time because Actions checks out the merge commit, so a PR adding a duplicate number sees both files. Fold in a `**Date:**`-line assertion here too rather than as its own gate.

[...]

Issue #944, opened 2026-08-03
Fix the ADR numbering collision and write the missing recap ADR


mglst:
## Question

ADR numbering has raced, and the fix has to land before any new ADR is written — several tickets under #901 want to create ADRs.

**The collision.** Ten code and test comments cite `ADR-0005`, all meaning *the recap decision* (no winner crowned, CO2 reported un-netted). No such file was ever written on any ref. Meanwhile two unmerged branches already own the numbers for unrelated subjects, both pushed to origin:

- `origin/docs/role-taxonomy-spec` → `docs/adr/0004-role-taxonomy-and-admin-authority.md` (#897/#898) — written, unmerged, **zero citations**. ADR-0004 is not missing.
- `origin/docs/934-adr-postgres-persistence` → `docs/adr/0005-game-instance-persistence-postgres.md`

So on merge, **all ten recap citations resolve to an ADR about Postgres** — actively misleading, not merely dangling.

**Root cause:** the ADR convention says "scan `docs/adr/` for the highest number and increment," which is race-prone by construction. Decide the replacement. Note two tickets under #901 also want new ADR numbers (typography, the error-handling wire contract), so whatever scheme is chosen must handle concurrent branches.

**Then write the recap ADR** on a free number and sweep all ten citations to match. The outline is drafted in `docs-policy-review/corpus-core.md` on branch `research/docs-policy-review`, sourced from #906 (which carries the complete rejected-options list), #907, and `schemas/recap.py`:

- **No winner crowned** — no rank, no medal on any row.
- **CO2 as two un-netted columns.** Netting was rejected as *gameable* (capture a pile, go net-negative, flatter a big emitter) and because it hides the produce/capture relationship. A normalised efficiency ratio was also rejected — it manufactures a contest the simulation is not running.
- Default `operating_income` descending, reframed as "most consequential first".
- Notability is per-column, client-derived, never minted.
- Ties break on `account_id` so re-mints are reproducible.
- The consequence a future reader will trip on: gross produced needed **no new sim metric** — it is the sum of positive `cumul_emissions` categories, because capture is stored as a negative `carbon_capture` entry (`player.py:542`). That is why `test_player_emissions.py` exists.

While here: ADR-0003 has one drift worth folding in — the fail-closed default at `server_config.py:33` (`signups_enabled = False`) is a real security decision recorded only in a Python docstring. And none of the three existing ADRs has a date field, despite the convention treating them as dated.

Part of #901.

[...]

Issue #943, opened 2026-08-03
Design the docs-integrity CI checks


mglst:
## Question

Policy point 6 from #904: mechanism over exhortation. Machine-checked facts did not drift in this repo; prose drifted uniformly. Design the checks, then decide which ship.

Four proposed, ~200 lines total. Prior art already in the repo: `scripts/check_wiki_links.py` does exactly this for wiki MDX.

1. **`docs:refs` — every backticked repo path and markdown link in `docs/` resolves.** Root-absolute `/…` links are errors. **Fails today on 21 dead paths and 8 broken links** (one confirmed: `architecture/real-time-sync.md` links `/docs/frontend/data-fetching.md`, which breaks on GitHub). Its real value is timing: it would have failed CI on `b47edf85` ("remove legacy frontend"), forcing the author to fix or delete `hooks.md` in the one PR where deleting was obviously right. Instead 790 lines survived four more months, still pointing newcomers at `usePlayerMoney.ts`, `usePlayerWorkers.ts`, `useWeather.ts` — none of which exist.
2. **Inverted anchors — every `docs/**.md` path and `§ Section` referenced *from code* resolves.** **Fails today**: `frontend/src/styles/global.css:15` cites `docs/frontend/tailwind-v4-theming.md`, which has never existed in the repo's history. This is the check that makes pointing code at a doc a commitment. Note 10 code citations target `static-serving-and-deployment.md` **by section anchor**, so anchors must be in scope, not just paths.
3. **Orphan detection** — every doc reachable from code, from `docs/README.md`, or is an ADR. Makes stub files structurally impossible. Today all three ADRs are orphans and `docs/README.md` does not mention `docs/adr/` at all — the index is a table of contents for the dead half. Decide whether `CONTRIBUTING.md` counts as a root.
4. **Pin the two cross-language contracts** — a shared `phase_vectors.json` fixture read by both `test_instance_config.py` and `instances.test.ts`; and a script cross-checking Python `invalidate_queries([...])` literals against `query-client.ts`. The latter would have caught the `["facilities","all"]` error that **all three** frontend docs still contain.

Prototype them against the current tree and report what each catches. A check that fires 21 times on day one needs a decided answer: fix all, or allowlist and burn down.

[...]

PR #938, opened 2026-08-02
docs(adr): game-instance persistence off pickle onto Postgres

mglst:
## Summary

Drafts ADR-0005, assembling the decisions settled while charting the
[Game-instance persistence wayfinder map](https://github.com/felixvonsamson/Energetica/issues/924)
into a single document: co-located Postgres+TimescaleDB (entities +
time-series), one Postgres server with one database per game instance,
Action-log/Save/Checkpoint/Replay coexistence (Postgres replaces the
Save/Checkpoint layer only), an ORM adoption with a SQLModel leaning,
new-instances-only migration, and a deferred local-dev-ergonomics
commitment.

Also folds in the codebase impact survey (#932) and the `ActiveFacility`
SQLModel prototype (#933) as Context/Prototype sections.

Concrete schemas, migration scripts, and the local-dev mechanism are
explicitly deferred to a follow-on spec map — this ADR is scoped to the
architectural decision only.

## Test plan

- [ ] Felix review & sign-off (#935)
- [ ] Yassir review & sign-off (#936)

<h3>Greptile Summary</h3>

This PR proposes replacing per-instance pickle persistence with a shared PostgreSQL/TimescaleDB server using one database per game instance, while retaining the Action log and replay model.

- Selects relational storage and an ORM, with SQLModel as the current preference.
- Limits migration to new game instances and defers schemas, tooling, checkpoint mechanics, and local-development setup.
- Documents consistency, session lifecycle, identity, nested-data, and test-fixture concerns for the follow-on specification.

<h3>Confidence Score: 5/5</h3>

The documentation-only PR appears safe to merge, with implementation-sensitive concerns clearly deferred to the follow-on specification.

The ADR consistently scopes the architectural decision, preserves the existing action-log and replay concepts, and explicitly records the major persistence, consistency, migration, checkpoint, and developer-environment work that remains.

<h3>Important Files Changed</h3>

| Filename | Overview |
|----------|----------|
| docs/adr/0005-game-instance-persistence-postgres.md | Adds a proposed architectural decision for migrating new game instances from whole-engine pickle persistence to per-instance PostgreSQL databases while recording implementation risks and deferred design work. |

<h3>Flowchart</h3>

```mermaid
%%{init: {'theme': 'neutral'}}%%
flowchart LR
  Actions[Action log] --> Replay[Replay]
  Replay --> DB[(Per-instance PostgreSQL database)]
  Entities[Entity tables] --> DB
  TimeSeries[TimescaleDB hypertables] --> DB
[...]

PR #981, opened 2026-08-09
docs(adr): write the missing recap ADR as 0005; date all ADRs

mglst:
Resolves the ADR numbering collision in #944.

## The problem

Ten code and test comments cite `ADR-0005` for the recap decision — no winner crowned, CO2 reported un-netted. **The file was never written on any ref.** Meanwhile `docs/934-adr-postgres-persistence` (PR #938) had already claimed `0005` for game-instance persistence on Postgres. On merge, all ten recap citations would have resolved to a document about Postgres — actively misleading rather than merely dangling.

Root cause: the ADR convention says "scan `docs/adr/` for the highest number and increment", which is race-prone by construction. Three parallel branches scanned a `main` whose highest was `0003`; two independently allocated `0004`/`0005` and a third cited `0005` from code without writing it.

## What this PR does

- **Writes `docs/adr/0005-recap-is-retrospective-not-a-scoreboard.md`.** Sourced from #906 (which carries the full rejected-options list), #907, and the docstrings in `energetica/schemas/recap.py`. The ten existing citations are already correct and need no sweep.
- **Adds a `Date:` line to all four ADRs.** The convention treated ADRs as dated; none of them were. Backfilled from first-commit dates.
- **ADR-0003: records that the signup toggle fails closed.** The `False` default, the `StrictBool` typing and the missing-file behaviour in `energetica/server_config.py` are a real security decision that lived only in a Python docstring.

## Why recap takes 0005 rather than a free number

Ten citations in merged, shipped code already say `0005`, and the recap decision (2026-07-22) predates the Postgres one (2026-08-02). Giving recap a free number would make correct code wrong so that an unmerged PR could keep a filename. **PR #938 should renumber to `0006`** — it has zero inbound citations.

## What is deliberately not here

No prose about the numbering rule. Per the docs policy ratified in #904, enforcement belongs to tooling, so the race is fixed by two CI checks landing with #979:

1. No two ADR files share a number — fires at PR time, since Actions checks out the merge commit.
2. Every `ADR-NNNN` token in tracked source resolves to a file in `docs/adr/`. This is the check that would have caught the original defect, and #943's ratified `docs:refs` grammar cannot see it (that grammar keys on `/`).

Check 2 is **red on `main` today, ten times**, and only goes green once this PR lands. So this PR blocks #979.

<h3>Greptile Summary</h3>

The PR adds the missing recap ADR, dates the existing ADRs, and aligns recap documentation with the implemented self-healing mint guard.
- Records the retrospective, non-competitive recap design as ADR-0005.
- Documents the fail-closed server-wide signup toggle in ADR-0003.
- Corrects recap lifecycle documentation to distinguish readable artifacts from bare file existence.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| docs/adr/0005-recap-is-retrospective-not-a-scoreboard.md | Adds the missing recap decision record and accurately documents `load_recap` as the mint-once and corruption-recovery guard. |
| docs/architecture/static-serving-and-deployment.md | Corrects the recap publication description to match the readable-artifact guard and self-healing behavior. |
| energetica/utils/recap.py | Updates module documentation to accurately describe the existing mint-once and corrupt-artifact recovery behavior. |
| docs/adr/0003-account-creation-decoupled-from-instance-access.md | Adds the ADR date and records the fail-closed signup configuration decision. |
| docs/adr/0001-action-log-stays-complete-fix-oom-on-read.md | Adds the historical decision date without changing the recorded behavior. |
| docs/adr/0002-server-wide-sso-shared-cookie.md | Adds the historical decision date without changing the recorded behavior. |

<sub>Reviews (2): Last reviewed commit: ["docs(adr): correct the recap mint-once g..."](https://github.com/felixvonsamson/energetica/commit/4aad14a24fda5b1ea1f1129f1350223eef36d45f) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51633688)</sub>


docs/adr/0005-recap-is-retrospective-not-a-scoreboard.md
```diff
@@ -0,0 +1,65 @@
...
+won", captured-CO2-as-brag-stat. The rest of G1 stands: tombstone model, frozen-photograph
+semantics, map snapshot, phase-derived visibility, static publication.
+**Decided in:** #906. **Applied by:** #907.
+
+## Context
+
+The recap is a write-once JSON tombstone minted at the `active → freeze` transition and
+published to `recaps/{slug}.json`, where it outlives the instance process
+(`energetica/utils/recap.py`, `mint_recap`, mint-once guarded by
+`instance_config.recap_exists`).
```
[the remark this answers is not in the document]
mglst: Correct, and it was worse than one line — thanks. Verified at `energetica/utils/recap.py:97`: the guard is `instance_config.load_recap(slug) is not None`, and `mint_recap_if_needed`'s own docstring says *"The guard is `load_recap` — not bare file-existence — so it also self-heals"*, with `test_recap.py:398-402` proving a corrupt recap re-mints.

Taken with two amendments to your suggestion and two extra sites:

1. **Your replacement kept one of my errors.** It attributes the mint-once guard to `mint_recap`, but `mint_recap` is the unconditional force-overwrite primitive — the guard lives in `mint_recap_if_needed`. Fixed to name the right function.
2. **The same false claim had two other homes.** `energetica/utils/recap.py`'s *module* docstring said "guarded by file existence" while the function docstring five lines below says the opposite, and `docs/architecture/static-serving-and-deployment.md` said it too — that is where the ADR draft inherited it from. Both corrected in `4aad14a2`, since leaving the original source intact would just have fed the error back in when that file is restructured under #946.

`tests/unit/test_recap.py` still passes (23).

docs/adr/0005-recap-is-retrospective-not-a-scoreboard.md
```diff
@@ -0,0 +1,66 @@
...
+G1 specified it as an income-ranked leaderboard that crowns a winner. Energetica is a
+systems-education game about the trade-offs of running a grid. A podium answers a question
+the simulation is not asking, and — worse — it tells the player which of the many things
+they can read off the table is *the* thing that mattered.
+
+## Decision
+
+The recap presents information matter-of-factly and lets the reader draw the conclusions.
+
+1. **No winner is crowned.** No global `rank`, no medal on any row. Enforced by
```
felixvonsamson, line 27: I'm not really sure where that comes from. Probably a missunderstandingbetween us. I think it is important to have a concept of winner even if there might be different categories. This is important for player motivation.
CO2 probably needs a better metric like CO2/MWh or something. 
The current recap page design already incorporate these changes, I don't know if we should update it in the ADR.

felixvonsamson (commented):
Agree with the changes but I have doubts about the content of ADR 0005. To me it is already outdated.

docs/adr/0005-recap-is-retrospective-not-a-scoreboard.md
mglst: adjusted slightly. now reads:
```md
1. **No overall winner is crowned.** Having per-category winners and rankings is fine.
```
`docs/adr/0005-recap-is-retrospective-not-a-scoreboard.md:27-28`

Issue #953, opened 2026-08-03
Ruff and vitest never run in CI; 82 lint errors are live outside energetica/


mglst:
Two gaps in CI that let checked-in code drift from the configured standards.

**Ruff is not in CI at all.** It runs in pre-commit only, scoped to `^energetica/`. Consequence today: **82 lint errors and 3 unformatted files** across `lobby/`, `tests/` and `scripts/` — none of which any hook or workflow looks at. Note Pyright already covers the repo at `standard`, so the Python story is inconsistent rather than absent: type checking is repo-wide, linting is one directory and advisory.

**Vitest is not in CI.** Four test files exist and never run.

Fix: add both to `.github/workflows/ci.yml`, and widen the Ruff scope to the whole repo. The 82 existing errors need clearing (or an explicit, shrinking allowlist) in the same change, or the job lands red.

This is a prerequisite for part of the docs-to-rules backlog in #942 — several of those items are Ruff rules (`I` for import order, `N` for naming, neither currently selected), and adding a rule to a linter that CI ignores accomplishes nothing.

Found while grading documented conventions against actual enforcement for #904.

Issue #946, opened 2026-08-03
Execute the ratified prune, restructure, and moves


mglst:
## Question

Nothing to decide — carry out the ratified dispositions. This is the "do" the map's Notes authorise.

Scope:

1. Delete, move, split and refresh every file per #941.
2. Rescue the passages the dry-runs quoted verbatim into their decided destinations — roughly 60 lines out of 3,521 in the frontend corpus. They do **not** consolidate: they land in three different places (two new ADRs, four code-comment or `CLAUDE.md` edits, one surviving doc). A single "frontend notes" file would recreate exactly the mixed-claim grab-bag the policy exists to prevent.
3. Preserve citation integrity. 10 code comments target `static-serving-and-deployment.md` **by section anchor**; 22 inbound references point at files marked for cutting, including 2 in out-of-scope `docs/agents/domain.md` and 1 in `architecture/real-time-sync.md`, which is itself a survivor. `CONTRIBUTING.md:222` links both style guides.
4. **Rewrite `docs/README.md` as a map, not a catalogue.** Today it devotes 16 bullets to `frontend/`, several pointing at stubs — including the literal line `[**Style Guide**](frontend/style-guide.md) - TODO` — while never mentioning `docs/adr/`, the artifact cited 56 times from code.
5. Update the embedded tree listing in `docs/agents/domain.md` to match. Its contents are out of scope but the listing describes the tree, so it goes stale on this change.
6. Fix `CLAUDE.md:19`, which still declares the deleted Jinja frontend as part of the stack.
7. Clean up two dead references to a deleted `typography-examples.tsx` (`docs/frontend/typography.md:139` and `frontend/src/routes/app/internal/typography.tsx:399`).

Write against the one exemplar in the tree: `docs/backend/deployment.md`, added 2026-08-03, is the **only file in the corpus needing zero edits** — near-pure procedure, one pointer out instead of restating the design.

Blocked by #941 and #945.

Part of #901.

mglst:
**Three reference-repair obligations added by #943.** The docs-integrity checks land in #979 immediately after this ticket, with no allowlist — so this ticket is what makes them green.

[...]

Issue #942, opened 2026-08-03
Decide which cut docs become lint rules or tests


mglst:
## Question

Policy point 4 from #904: every cut doc leaves a test behind. Decide **which** of the enforceable-but-unenforced claims become tooling, in what order, and which are dropped instead.

76 prescriptive claims were graded against the real tooling (`docs-policy-review/enforcement.md` on branch `research/docs-policy-review`): 18 already enforced, 28 generic advice, 15 genuinely unenforceable, **15 enforceable but unenforced**. The last group is the backlog.

Four of them replace prose the codebase **already contradicts at scale** — a convention with no checker is one nobody follows:

| prose claim | reality |
|---|---|
| filenames are `PascalCase.tsx` | **0 of 133** files comply — all kebab-case |
| enum values are `SCREAMING_CASE` | **39 of 48** violate |
| `tan-green` / `brand-green` come from `tailwind.config` | no such file (Tailwind v4), both tokens **undefined despite 81 usages** |
| "always include `dark:` variants" | backwards — `global.css:340` swaps semantic tokens under `:root.dark` |

Candidates, roughly by value over effort:

- **Add Ruff to CI at all** — prerequisite for the Python items, ~82 errors to clear first (see the separate CI issue).
- **Tailwind token guard** — highest bug yield, catches ~81 live defects today.
- **Filename convention test** (~20 lines).
- **Error-envelope contract test against both apps** (~30 lines) — the one thing `api-types-freshness` structurally cannot see.
- **`GameExceptionType` value test** with a shrinkable 39-entry legacy allowlist.
- **Ruff `"I"` (import order) and `"N"` (naming)** — currently unselected, so `backend/style-guide.md`'s import-order rule and its entire naming table have zero enforcement.
- **`no-inferrable-types`**, dummy-data guard, `commitlint` (the prose admits it is "enforced *culturally*").

Decide: which land in this effort, which get filed as follow-ups, which are dropped as not worth a checker. A rule nobody will fix when it fires is worse than no rule.

Prior art to match: `scripts/guard-no-raw-html.ts` and `tests/unit/test_module_boundary.py` — both in CI, both with teeth.

Part of #901.

mglst:
Handoff from #941 (per-file dispositions), so this ticket knows what has already been routed to it. Four candidates fell out of the disposition pass, in descending order of value:

[...]

Issue #904, opened 2026-07-19
Decide the docs/ curation plan — per-file keep/cut/move + target tree


mglst:
## Question

Given the audit (#903) and Felix's ratification (#902), produce the **curation plan**: the final disposition of every file under `docs/` plus the target tree. This is the artifact the execution pass follows.

Decide:
- **Confirm the 15 cut candidates** — 7 dead-stubs (`backend/databse.md`, `backend/game-engine-startup.md`, `game-logic/charts.md`, `frontend/{routing,state-management,component-library,style-guide}.md`) + 8 generic-noise (`backend/style-guide.md`, `frontend/{design-patterns,hooks,animations,styling,best-practices}.md`, and the two flagged `agents/*` — but see scope note below).
- **Keep-as-is vs keep-but-refresh** for the ~22 onboarding-core + architecture-of-record files (all 3 ADRs verified current).
- **Self-hosting** (2 files) — subset page under `getting-started/`, or drop.
- **Address the five cross-cutting staleness drivers** the audit found: (1) Jinja→React migration complete (`energetica/templates/` gone); (2) kebab-case file naming repo-wide; (3) auth moved to the lobby + shared `energetica_session` cookie; (4) Tailwind v4 has no `tailwind.config`; (5) `bun run generate-types` not `bun generate-types`.
- **Target tree** consistent with `docs/agents/domain.md` (root `CONTEXT.md` + `docs/adr/` + topic tree).

**Scope note:** `docs/agents/*` are skills config written by `setup-matt-pocock-skills`, NOT onboarding docs. Do not cut them; they are out of scope.

Audit findings: #903 comment + branch `research/docs-audit`.

Part of #901. Blocked by #902.

mglst:
**Resolution.** The ticket asked for a per-file keep/cut/move table. The session found that framing was wrong, so what it produced instead is the **policy that decides dispositions**, plus a redrawn destination. The per-file table becomes mechanical once the policy is settled, and is now its own ticket.

## What changed the framing

The tree is longest where knowledge is free and shortest where it is scarce. `docs/frontend/` is 58% of the tree and mostly third-party tutorial — 722 lines of TanStack Query, 309 of Tailwind. `docs/game-logic/`, covering the electricity-market simulation behind the 3,674 hardest lines in the repo, is **20 lines**: one stub, plus a vocabulary list with literal `<>` where the definitions belong.

[...]

PR #982, opened 2026-08-09
chore(ci): land the filename and dummy-data guards (#971)

### LINKED ISSUE #971: Land the filename-convention and dummy-data guards
## Question

Two guards that collide on one file, so they ship together.

**Filename convention test.** Every file under `frontend/src` matches `^[a-z0-9]+(-[a-z0-9]+)*\.(ts|tsx)$`, with an allowlist for `routeTree*.gen.ts`. All 133 files already comply — the prose in `backend/style-guide.md` and `frontend/best-practices.md` asserts `PascalCase.tsx`, which is the opposite of universal reality. This test pins the reality and deletes the claim. Roughly 20 lines.

**Dummy-data guard.** Forbid importing `@/data/dummy*` outside an `import.meta.env.DEV` branch. Clone `scripts/guard-no-raw-html.ts`.

They collide on `frontend/src/data/dummyFacilities.ts`: it is the one file that would fail the filename test, and it is also the module that currently ships in the production bundle behind a `SHOW_DUMMY_DATA` constant. Decide its fate — delete, rename and gate, or move — and both guards go green.

Part of #901. Decided in #942 (backlog items 2 and 10).

mglst:
Closes #971. Part of the [docs reallocation map (#901)](https://github.com/felixvonsamson/Energetica/issues/901); decided in #942 (backlog items 2 and 10).

Two guards that collided on one file, so they ship together.

## The guards

**`scripts/guard-filenames.ts`** pins the kebab-case naming that all 321 tracked `.ts` / `.tsx` files under `frontend/src` already follow. The check's real content is its carve-outs:

- Generated names are exempt by suffix (`*.gen.ts`, `*.generated.ts`) — `routeTree.gen.ts` and `api.generated.ts` are named by the tools that write them, and exempting by suffix means a new route tree needs no allowlist edit.
- `.d.ts` and `.test.ts` / `.test.tsx` are stripped before the stem is judged, so `single-origin.test.ts` isn't failed for a dot it's required to have.
- Inside the three route trees the stem is a TanStack file-based route path, not a plain name. `runs.$slug.recap.tsx` is three segments, `$slug.tsx` is a parameter, `__root.tsx` is the root layout. Each dot-separated segment is checked on its own, so `routes/app/DashboardCard.tsx` still fails.

**`scripts/guard-no-dummy-data.ts`** forbids importing a module whose path names it as fixture data, except from a test file or from `routes/app/internal/`. The rule is on the **import edge**, not the file: an unimported fixture is tree-shaken away and harms nothing, while an imported one is in the bundle no matter how the call site is guarded.

The word list is deliberately short — `dummy`, `mock`, `fake`, `fixture`, `stub` and plurals. `sample`, `seed` and `placeholder` were tried and dropped: a chart sample rate and an RNG seed are ordinary vocabulary here, so a firing on `sample-rate.ts` would get suppressed rather than fixed, which is the kind of rule #942 ruled out.

## The file they collided on is deleted

`frontend/src/data/dummyFacilities.ts` was 297 lines of hand-written facility records that `routes/app/facilities/manage.tsx` imported behind `const SHOW_DUMMY_DATA = false as boolean` — an assertion whose only purpose is to defeat type narrowing, and which therefore also defeats dead-code elimination. Both the toggle and the 297 lines shipped in every production bundle.

Deleted rather than renamed and gated: its stated purpose was "showcasing all asset colors", and `routes/app/internal/colors.tsx` already does that across 32 facility types to the dummy module's 21 — a strict superset, verified type by type. `manage.tsx` now just renders `facilities`.

## Wiring and prose

Both guards run as steps in the existing `frontend-checks` CI job, next to the raw-HTML guard, and are exposed as `bun run guard:filenames` / `bun run guard:no-dummy-data`. Neither uses vitest, so neither waits on #953.

Per #904 — every cut claim leaves a test behind — the prose these replace is deleted. `docs/frontend/best-practices.md` asserted components are `PascalCase.tsx`, hooks `usePlayerMoney.ts` and utils `camelCase.ts`; not one of those described a single file in the repo. `docs/backend/style-guide.md`'s module row now says kebab-case. Both files are otherwise left for #946 to dispose of.

## Verification

Both guards were run against the real tree (green) and against deliberately planted violations (red), including confirming the `/app/internal/` carve-out lets its importer through. `typecheck`, `lint`, `format:check` and `build:app` all pass.

<h3>Greptile Summary</h3>

The PR removes production dummy facility data and adds CI guards enforcing frontend filename conventions and preventing fixture-data imports from player-facing code.

- Deletes the dummy facility module and simplifies the facility-management route to use API data directly.
- Adds Bun scripts and frontend CI steps for filename and fixture-import validation.
- Updates frontend and backend naming documentation to match the enforced kebab-case convention.

The PR appears safe to merge with no concrete blocking or non-blocking defects identified.

The new CI guards are correctly invoked against tracked frontend sources, and the dummy-data removal preserves the existing live-data loading, error, and rendering paths.

| Filename | Overview |
|----------|----------|
| scripts/guard-filenames.ts | Adds a tracked-source filename guard with exemptions for generated files, declarations, tests, and current TanStack route naming forms; no actionable defect found. |
| scripts/guard-no-dummy-data.ts | Adds a repository-aware import-edge scanner that rejects fixture-named local modules outside tests and the internal design-system routes; no actionable defect found. |
| frontend/src/routes/app/facilities/manage.tsx | Removes the disabled dummy-data branch and consistently renders loading, error, and live facility data states. |
| .github/workflows/ci.yml | Wires both new guards into the existing frontend-checks job with the correct package working directory. |
| frontend/package.json | Exposes both guard scripts through Bun without changing dependencies or build behavior. |

<sub>Reviews (1): Last reviewed commit: ["chore(ci): land the filename and dummy-d..."](https://github.com/felixvonsamson/energetica/commit/f1a3b5f3378c4b8a207089af8a75efb2a9a00849) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51653192)</sub>


scripts/guard-no-dummy-data.ts
```diff
@@ -0,0 +1,155 @@
...
+ * hyphen-separated words in a filename stem and against whole path segments,
+ * never as substrings.
+ *
+ * The list is deliberately short. `sample`, `seed` and `placeholder` were tried
+ * and dropped: they are ordinary technical vocabulary here — a chart sample
+ * rate, an RNG seed, an input placeholder — so a firing on `sample-rate.ts`
+ * would get suppressed rather than fixed, which is exactly the kind of rule
+ * #942 ruled out. What remains has no innocent reading as a module name.
+ */
+const FIXTURE_WORDS = new Set([
```
felixvonsamson, line 57: That looks a bit unnecessary but ok

felixvonsamson (approved):

scripts/guard-no-dummy-data.ts
mglst: I agree - most comments are honestly useless if not actively harmful by being noisy and verbose. But that's a larger issue.

PR #1044, opened 2026-09-02
Add scripts/shift_start_date.py to pause an instance over a break

### LINKED ISSUE #1024: Pause the ZHAW instance over the holiday break without simulating the downtime
## Context

One-off use case for the ZHAW instance (project with Christian): the run needs to stop for a
holiday break of a few weeks, then resume afterwards without the game having simulated that
downtime as elapsed game time.

## Why this needs a manual step

The sim's "epoch" is `engine.start_date`, persisted in `instance/engine_data.pck`. Every tick,
the scheduler computes how many ticks *should* have run by now as:

```python
total_t_target = (time.time() - engine.start_date.timestamp()) / engine.clock_time
```

If the server is simply stopped and restarted after the break with nothing else changed, it will
see the full real-world gap since `start_date` and rapid-fire every missed tick to catch up —
i.e. it simulates the whole holiday. Editing `start_date` in the instance's config file does
**not** help: config's `start_date` is only ever consulted by `init_instance()`, which runs once
for a brand-new instance (no `instance/engine_data.pck` yet). For an instance already in
progress, restart always takes the `engine.load()` path and restores `start_date` straight from
the pickle, ignoring the config entirely. There's also no built-in pause/resume in the instance
lifecycle — `announced → active → freeze → ended` is one-directional; `freeze` is a permanent
"run over, mint the recap" state, not a resumable pause.

## The plan

1. Stop the server normally for the holiday break (or, more safely, first run
   `engine.save_checkpoint()`/back up `instance/` before stopping).
2. While stopped, shift the persisted `engine.start_date` forward by exactly the real downtime,
   directly in `instance/engine_data.pck`:

   ```python
   import pickle
   from datetime import timedelta

   with open("instance/engine_data.pck", "rb") as f:
       data = pickle.load(f)

   data["start_date"] += timedelta(days=X)  # X = the real downtime, in whole days

   with open("instance/engine_data.pck", "wb") as f:
       pickle.dump(data, f)
   ```

3. Restart the server. It resumes as if no time had passed — no catch-up burst.

### Notes / constraints

- Shift by a **whole number of days** (or any multiple of `clock_time`). All supported
  `clock_time` values (60, 30, 20, 15, 12, 10, 6, 5, 4, 3, 2, 1) divide 86400 evenly, so a
  whole-day shift preserves the tick alignment the daily-quiz trigger relies on
  (`tick_execution.py`, `% 86400 == 9*3600`).
- Do the edit only while the process is fully stopped — `engine.load()` raises if any file
  under `instance/data/**` looks newer than `engine_data.pck`, and editing the pickle bumps its
  mtime, which is what we want (it must stay the newest file).
- Back up `instance/` (or run `engine.save_checkpoint()`) before editing, in case the shift value
  is wrong.

### Verified: this only affects tick catch-up pacing, nothing else

Traced every consumer of both `total_t` and real wall-clock time before committing to this plan:

- All gameplay state (ongoing projects/upgrades, resource shipments/market arrivals, weather/
  climate, daily-question rotation, tech levels) is keyed off `total_t` (simulated ticks) and
  `delta_t`, never off `engine.start_date` — unaffected by the shift.
- The frontend doesn't reconstruct an absolute in-game date from `start_date` either: tick
  numbers are pushed live over the socket, and every "time remaining" display is a pure delta
  (`(end_tick − current_tick) × seconds_per_tick`). `start_date` is fetched via `/game/engine`
  but not currently consumed by any component.
- Things that are genuinely wall-clock-anchored (chat message timestamps, `Player.created_at` /
  `last_connection`, auth/session bookkeeping) don't reference `engine.start_date` at all, so
  they correctly keep advancing through the real holiday gap — that's desired, not a side effect
  to work around.
- Instance lifecycle boundaries (`starts_at`/`freeze_at`/`ended_at` in `instance.json`) live
  entirely separately from the pickle and are untouched by this edit.

## Optional follow-up

Could turn the snippet above into a proper `scripts/shift_start_date.py` with validation (reject
non-multiple-of-`clock_time` shifts, print before/after, refuse to run while the server is up) —
not needed unless this becomes a recurring need beyond this one-off.

felixvonsamson:
Closes #1024.

Adds the script the issue's follow-up asked for (endorsed in the comment thread): shift the persisted `engine.start_date` in `instance/engine_data.pck` by a whole number of days, so a stopped-and-restarted instance doesn't rapid-fire every missed tick to catch up on real downtime (e.g. a holiday break). Editing the instance config's `start_date` doesn't work for this — it's only read by `init_instance()` on a brand-new instance, never by `engine.load()`, which is the path a restart of an already-running instance always takes.

**What it does**
- Validates the shift is nonzero and a whole multiple of the instance's `clock_time` — keeps the daily-quiz trigger's tick alignment intact (`% 86400 == 9*3600` in `tick_execution.py`).
- Refuses to run if it finds a live `main.py` process whose cwd matches the instance (best-effort `/proc` scan; `--force` to override).
- Backs up the pickle before writing (`<pickle>.bak-<timestamp>`), unless `--no-backup`.
- Supports `--dry-run` to check the math first.

**Testing**
- `tests/unit/test_shift_start_date.py` — 21 tests on the two pure-logic seams (`shift_start_date`, `find_running_instance_pid`), matching this repo's existing script-test convention (pickle I/O/CLI glue stays in `main()`, untested, per `test_migrate_drop_user.py`'s precedent).
- Full backend suite passes (362 tests).
- Manually smoke-tested end to end (dry-run, real run + backup, zero-days rejection, missing-pickle error) against a synthetic pickle.
- Ran through `/code-review`; the one finding (unused import in the test file) is fixed.

**Docs**
- `docs/backend/incident-recovery.md` gets a new "Pausing an instance without simulating downtime" section with the runbook.

<h3>Greptile Summary</h3>

This PR adds an operator-facing script and runbook for moving a stopped instance's persisted simulation epoch across a real-world break.
- Validates whole-tick day shifts and offers dry-run and backup modes.
- Attempts to detect a live instance before modifying its engine pickle.
- Adds focused unit coverage for shift validation and process detection.

The live-pickle replacement should be made atomic before merging so an interrupted maintenance command cannot prevent the instance from restarting.

The script truncates the only pickle used by normal startup before the replacement serialization succeeds, and startup has no automatic fallback to the backup it creates.

**Files Needing Attention:** scripts/shift_start_date.py

| Filename | Overview |
|----------|----------|
| scripts/shift_start_date.py | Adds the epoch-shifting CLI and safety checks, but writes the critical live pickle non-atomically. |
| tests/unit/test_shift_start_date.py | Covers shift validation and process detection thoroughly, while leaving the destructive write path untested. |
| docs/backend/incident-recovery.md | Documents the intended stop-shift-restart operational procedure and required service user. |

<h3>Flowchart</h3>

```mermaid
flowchart TD
  Stop[Stop instance service] --> Guard[Check for matching live main.py]
  Guard --> Load[Load engine_data.pck]
  Load --> Shift[Shift persisted start_date]
  Shift --> Backup[Copy original to timestamped backup]
  Backup --> Write[Write modified live pickle]
  Write --> Start[Restart instance service]
  Write -. interruption or I/O failure .-> Partial[Partial live pickle]
  Partial --> Failure[Restart fails during pickle.load]
```

<sub>Reviews (1): Last reviewed commit: ["Add scripts/shift\_start\_date.py to pause..."](https://github.com/felixvonsamson/energetica/commit/245c244f0399ab4aea299867f77120be01492e2a) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=59692612)</sub>

**Context used:**

- Knowledge Base — [Game data and persistence](https://app.greptile.com/energetica-org-2/-/custom-context/knowledge-base/felixvonsamson/energetica/-/docs/game-data-and-persistence.md)
- Knowledge Base — [Deployment and environment operations](https://app.greptile.com/energetica-org-2/-/custom-context/knowledge-base/felixvonsamson/energetica/-/docs/deployment-and-environment-operations.md)

### DIFF
--- a/docs/backend/incident-recovery.md
+++ b/docs/backend/incident-recovery.md
@@ -100,3 +100,16 @@ Persistence runs at two cadences (see `utils/tick_execution.py`):
 - **`save_checkpoint()` — every 6 hours.** Writes `checkpoints/last_checkpoint.tar.gz`, a tarball of the entire `instance/` directory (including a fresh `engine_data.pck` and all data files) at a consistent point in time. This is the unit of disaster recovery.
 
 Check `ls -la checkpoints/` to see the current checkpoint's age. Because the checkpoint is only every 6 h, recovery relies on replaying the actions log forward from the checkpoint tick — which is why the log must never be truncated below a tick you might need to restore from (see `docs/adr/0001-action-log-stays-complete-fix-oom-on-read.md`).
+
+## Pausing an instance without simulating downtime
+
+There's no built-in pause/resume: `announced → active → freeze → ended` is one-directional, and `freeze` mints the recap rather than allowing a resumable pause. If an instance needs to stop for a real-world break (e.g. a holiday) and resume afterwards without simulating that gap as elapsed game time, shift the persisted epoch forward by the downtime instead (issue #1024):
+
+```bash
+cd /var/www/energetica-{slug}
+sudo systemctl stop energetica-{slug}
+sudo -u energetica .venv/bin/python scripts/shift_start_date.py --pickle instance/engine_data.pck --days <N>
+sudo systemctl start energetica-{slug}
+```
+
+`<N>` is the real downtime in whole days (or any other whole multiple of the instance's `clock_time` — every supported `clock_time` divides a day evenly, so whole days are always safe). The script refuses to run against an instance it can still see running, backs up the pickle before writing, and prints the old/new `start_date`; pass `--dry-run` first to check the math. Editing `start_date` in the instance's config file does **not** work here — it's read only by `init_instance()`, never by `engine.load()`, which is the path a restart of an already-running instance always takes.
--- a/scripts/shift_start_date.py
+++ b/scripts/shift_start_date.py
@@ -7,10 +7,7 @@ tick, the scheduler computes how many ticks *should* have run by now as
 ``(time.time() - engine.start_date.timestamp()) / engine.clock_time``. If an instance is simply
 stopped and restarted after a break (holiday, maintenance window, ...) with nothing else changed,
 it sees the full real-world gap since ``start_date`` and rapid-fires every missed tick to catch
-up — i.e. it simulates the whole break. Editing ``start_date`` in the instance's *config* file
-does not help: config's ``start_date`` is only ever read by ``init_instance()``, which runs once
-for a brand-new instance. An instance already in progress always restores ``start_date`` straight
-from the pickle on restart (``GameEngine.load()``), ignoring the config entirely.
+up — i.e. it simulates the whole break.
 
 This script shifts the persisted ``start_date`` forward (or back) by a whole number of days, so
 the instance resumes as if no time had passed — no catch-up burst. Shifting by a whole day (or
@@ -43,6 +40,7 @@ Usage:
 from __future__ import annotations
 
 import argparse
+import os
 import pickle
 import shutil
 import sys
@@ -157,8 +155,13 @@ def main() -> int:
         shutil.copy2(args.pickle, backup_path)
         print(f"Backed up original pickle to {backup_path}")
 
-    with args.pickle.open("wb") as f:
+    # Write to a temp sibling and atomically replace the live pickle, rather than truncating it
+    # in place: if the process is interrupted or hits an I/O error mid-dump, the original file
+    # must stay intact instead of being left half-written and unreadable on next startup.
+    tmp_path = args.pickle.with_name(f"{args.pickle.name}.tmp-{os.getpid()}")
+    with tmp_path.open("wb") as f:
         pickle.dump(engine_state, f)
+    tmp_path.replace(args.pickle)
     print(f"Saved {args.pickle}")
     return 0
 
--- a/scripts/shift_start_date.py
+++ b/scripts/shift_start_date.py
@@ -9,11 +9,15 @@ stopped and restarted after a break (holiday, maintenance window, ...) with noth
 it sees the full real-world gap since ``start_date`` and rapid-fires every missed tick to catch
 up — i.e. it simulates the whole break.
 
-This script shifts the persisted ``start_date`` forward (or back) by a whole number of days, so
-the instance resumes as if no time had passed — no catch-up burst. Shifting by a whole day (or
-any other multiple of the instance's ``clock_time``) keeps tick alignment intact — in particular
-the daily-quiz trigger (``utils/tick_execution.py``, ``% 86400 == 9*3600``), which depends on
-``start_date`` landing on the same second-of-day it did before the shift.
+This script shifts the persisted ``start_date`` forward (or back), so the instance resumes as if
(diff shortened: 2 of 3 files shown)


docs/backend/incident-recovery.md
```diff
@@ -100,3 +100,16 @@ Persistence runs at two cadences (see `utils/tick_execution.py`):
...
 Check `ls -la checkpoints/` to see the current checkpoint's age. Because the checkpoint is only every 6 h, recovery relies on replaying the actions log forward from the checkpoint tick — which is why the log must never be truncated below a tick you might need to restore from (see `docs/adr/0001-action-log-stays-complete-fix-oom-on-read.md`).
+
+## Pausing an instance without simulating downtime
+
+There's no built-in pause/resume: `announced → active → freeze → ended` is one-directional, and `freeze` mints the recap rather than allowing a resumable pause. If an instance needs to stop for a real-world break (e.g. a holiday) and resume afterwards without simulating that gap as elapsed game time, shift the persisted epoch forward by the downtime instead (issue #1024):
+
+```bash
+cd /var/www/energetica-{slug}
+sudo systemctl stop energetica-{slug}
+sudo -u energetica .venv/bin/python scripts/shift_start_date.py --pickle instance/engine_data.pck --days <N>
```
mglst, line 111: Since we already `cd` into the relevant instance directory, I would think the `shift_start_date.py` script should infer the correct instance folder and pickle file on its own, rather than needing to pass it along as an option.

docs/backend/incident-recovery.md
```diff
@@ -100,3 +100,16 @@ Persistence runs at two cadences (see `utils/tick_execution.py`):
...
+sudo systemctl start energetica-{slug}
+```
+
+`<N>` is the real downtime in whole days (or any other whole multiple of the instance's `clock_time` — every supported `clock_time` divides a day evenly, so whole days are always safe). The script refuses to run against an instance it can still see running, backs up the pickle before writing, and prints the old/new `start_date`; pass `--dry-run` first to check the math. Editing `start_date` in the instance's config file does **not** work here — it's read only by `init_instance()`, never by `engine.load()`, which is the path a restart of an already-running instance always takes.
```
mglst, line 115: > or any other whole multiple of the instance's `clock_time`

This seems wrong. What would such an invocation look like? Because the `--days` option surely should only accept days, not ticks. Unless there is a `--ticks` option, in which case the parenthetical in this paragraph should explicitly mention the correct syntax.

mglst (approved):
Seems fine, looks sensible. Left minor comments on documentation.

docs/backend/incident-recovery.md
felixvonsamson: Would it not be cleaner to not have to cd to the relevant instance folder ?

docs/backend/incident-recovery.md
felixvonsamson: Last commit adds a --tick

PR #1066, opened 2026-09-12
Reorganize scripts/ by execution target, fix shebang, delete dead migrations

mglst:
## Summary

Fixes the production incident where `./scripts/grant-facilitator.py` failed with `ModuleNotFoundError: No module named 'apscheduler'` — the shebang (`#!/usr/bin/env python3`) resolved to system Python instead of the venv, because `scripts/` mixed dev-machine tools, lobby-only/instance-only ops scripts, and completed one-time migrations with no indication of where each is meant to run or how it gets there.

- **Split `scripts/` by deploy target**: `scripts/lobby/` (`grant-facilitator.py`, `whitelist-run.py` — touch the shared `accounts.db`, not any single instance) and `scripts/instance/` (`export_instance_to_csv.py`). `deploy-lobby.sh`/`deploy-instance.sh` now rsync only their own subset, flattened with no subfolder on the server, instead of blanket-copying all of `scripts/` (which also stops `scripts/dev/`, `scripts/lib/`, and the `*.ts` CI guards from uselessly reaching the server).
- **Relative shebang** (`#!.venv/bin/python`) on the three ops scripts instead of `#!/usr/bin/env python3` — they're always invoked from a directory where a venv sits right next to them (lobby dir or instance dir), the same convention the `.sh` scripts already rely on.
- **Delete three completed one-time migrations** (`backfill-instance-membership.py`, `migrate-drop-user.py`, `migrate-to-server-accounts.py`) and their tests. Updated the one live-code reference (`energetica/utils/misc.py`) to a TODO instead of pointing at a deleted script, and trimmed the now-orphaned runbook sections in `static-serving-and-deployment.md` and `incident-recovery.md`.
- **Move dev-only scripts** (`check_wiki_links.py`, `coin_generator.py`, `simplified_logo_generator.py`) into `scripts/dev/`, out of the sysadmin's day-to-day path.
- **Remove `DEPLOY_USER`** (env var + `--user`/`--deploy-user` flags) across all 9 call sites — never once set to anything but `"deploy"` in this project's history.
- **Add `scripts/push-bootstrap.sh`**: `scripts/infra/*.sh` can't be shipped by the normal deploy rsync (Option A — no server-side git checkout to rsync into before the first instance/lobby exists), so this flattens them into `/tmp` on the server, replacing manual scp.

## Test plan

- [x] `bash -n` on every edited `.sh` file
- [x] `python -m py_compile` on every moved/edited `.py` file
- [x] Full `tests/unit` suite passes (222 passed)
- [x] `bun run format` / `bun run lint` clean (no frontend files touched)
- [x] Grepped repo for stale references to old script paths, `DEPLOY_USER`, `--deploy-user`/`--user <ssh-user>` — none remain outside historical ADRs
- [ ] Not yet tested against a real deploy — `deploy-instance.sh`/`deploy-lobby.sh` changes are production-facing; recommend a dry-run or staging deploy before merging

<h2><a href="https://app.greptile.com/api/retrigger?id=63504455"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=1"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=1">Retrigger</picture></a>Confidence Score: 4/5</h2>

The PR is not safe to merge until a recovery mechanism remains available for memberships lost after a successful local settlement.

<h2>Findings</h2>

1. P1&nbsp;**Backfill recovery is removed** <a href="https://github.com/felixvonsamson/Energetica/pull/1066#discussion_r3997880068">▶</a>

<h3>Summary</h3>

- Instance deployments now ship only instance-scoped tools, while lobby deployments ship shared-account administration tools.
- Bootstrap infrastructure is transferred separately to `/tmp` for initial provisioning.
- One retired migration is still required as recovery for a reachable best-effort membership-write failure, so deleting it leaves settled players without a repair path.

<h3>Diagram</h3>

```mermaid
flowchart LR
  Repo[Local repository] --> Bootstrap[push-bootstrap.sh]
  Bootstrap --> Tmp[Server /tmp setup files]
  Tmp --> Provision[Provision lobby and instances]
  Repo --> LobbyDeploy[deploy-lobby.sh]
  LobbyDeploy --> Lobby[Lobby app root]
  LobbyDeploy --> LobbyScripts[scripts/grant-facilitator.py<br/>scripts/whitelist-run.py]
  Repo --> InstanceDeploy[deploy-instance.sh]
  InstanceDeploy --> Instance[Instance app root]
  InstanceDeploy --> InstanceScripts[scripts/export_instance_to_csv.py]
  Instance --> Settle[Create local Player]
  Settle --> Membership[Best-effort shared membership write]
  Membership -->|write succeeds| Discovery[Personal run discovery and private re-entry]
  Membership -->|write fails| Missing[Missing membership with no backfill]
```

<sub>Reviews (1) · Last reviewed commit: ["refactor(scripts): reorganize by executi..."](https://github.com/felixvonsamson/energetica/commit/93e1c4d40b5755f75dc111b58520dcd8fc78b4b3)</sub>


docs/architecture/static-serving-and-deployment.md
```diff
@@ -503,20 +503,13 @@ Scripts accept all inputs via arguments or env vars and support `--yes` to suppr
 
 No service restart — landing is pure static. `instances.json` and `instances/` are not touched — they are owned by the instance backends and must not be overwritten.
 
-### `migrate-to-server-accounts.py` flow
+### `migrate-to-server-accounts.py` flow (retired)
```
felixvonsamson, line 506: what is the added value of keeping this ?

docs/backend/deployment.md
```diff
@@ -11,15 +11,26 @@ the server.
 
 ## Initial Setup (One-Time)
 
-On the VPS, as root:
+`scripts/infra/*` (the `setup-*.sh` scripts below) are never shipped by the normal deploy
+scripts — Option A means there is no git checkout on the server for anything to rsync into
```
felixvonsamson, line 15: option A ??

docs/backend/deployment.md
```diff
@@ -80,12 +98,13 @@ the change up on the next request — no restart.
...
-`deploy-instance.sh`: `--yes` (skip confirm), `--skip-build`, `--skip-deps`, `--user <ssh-user>`.
-`deploy-landing.sh`: `--yes`, `--skip-build`, `--user <ssh-user>`.
-`deploy-lobby.sh`: `--yes`, `--skip-build`, `--skip-deps`, `--user <ssh-user>`.
+`deploy-instance.sh`: `--yes` (skip confirm), `--skip-build`, `--skip-deps`.
+`deploy-landing.sh`: `--yes`, `--skip-build`.
+`deploy-lobby.sh`: `--yes`, `--skip-build`, `--skip-deps`.
 
-All inputs also accept env vars (`DEPLOY_HOST`, `DEPLOY_USER`, `DEPLOY_DOMAIN`), so the
-scripts run unattended from CI.
+The SSH user is always `deploy` (hardcoded — never varied across this project's history).
```
felixvonsamson, line 110: Is that realy needed ? It makes sense now but I feel like this will be very confusing in the future

scripts/infra/setup-instance.sh
```diff
@@ -276,5 +277,5 @@ echo "  ./scripts/deploy-instance.sh --server <ssh-host> --instance $INSTANCE --
 echo
 echo "For a private/unadvertised instance, edit the policy before first login:"
 echo "  sudo \$EDITOR $CONFIG_DIR/instance.json   # set advertised/access.policy"
-echo "Then grow its roster (facilitator UI, or from the shell):"
-echo "  python scripts/whitelist-run.py $INSTANCE add <username> [<username> ...]"
+echo "Then grow its roster (facilitator UI, or from the shell, from the lobby directory):"
```
felixvonsamson, line 280: I would add the step "grant facilitator access" in the final instructions. I think there will almost always be a facilitator with a facilitator UI.

felixvonsamson (approved):
nice

felixvonsamson:
Notes of descisions:
- membership to an instance should not be granted at settle but when a player clicks "join Run" on the public run in the list on the lobby. 
- The "advertised" property of an instance can be removed. A public run is advertised and a private run is not.

src/energetica/utils/misc.py
```diff
@@ -267,10 +267,10 @@ def initialize_player(account: Account, tile: HexTile) -> Player:
...
     # Settling is what makes this run appear under the account's "your runs" in the lobby and the
     # in-run switcher. No-op without a slug (dev / unconfigured), where there is no lobby anyway.
-    # Best-effort, like instance_config.publish: this runs after the irreversible in-memory settle
-    # above, so a DB failure (e.g. SQLITE_BUSY on the shared accounts.db) must never propagate and
-    # fail an otherwise-successful settle — that would leave the player wedged (settled in-engine
-    # but the request 500s). A missing row is recoverable via scripts/backfill-instance-membership.py.
+    # TODO: a DB failure here (e.g. SQLITE_BUSY on the shared accounts.db) is currently swallowed
+    # so it doesn't 500 an otherwise-successful settle, but silently dropping the membership row
+    # with no alerting and no recovery path is not a validated design — revisit whether this
+    # should surface loudly instead (see discussion on scripts/ cleanup PR).
```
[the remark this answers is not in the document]
mglst: Good catch — you're right that deleting the migration removed the recovery path without replacing it with anything. Rather than restore the script, we fixed the actual gap: `initialize_player` now writes the `accounts.db` membership row *before* any of the in-memory engine mutation (tile claim, `Player` creation, chat join, facility, rolling history), and that write is left uncaught.

A transient DB failure (`SQLITE_BUSY` or similar) now aborts the whole settle before anything happens — no tile claimed, no orphaned `Player`, nothing to reconstruct — and surfaces as a plain 500 the client can safely retry. Previously the DB write happened *last*, after all that engine state was already mutated, which is why the exception had to be swallowed (a mid-mutation 500 would have left the player wedged with a claimed tile and no way to retry) — and why a recovery script was needed at all. Moving the write earlier removes the need for recovery rather than reintroducing it. See 069843ea.

docs/backend/deployment.md
mglst: I think there are good reasons _not_ to use `root` for deployment scripts. It causes permissions issues and, as I understand, is not considered 'best practice' since indeed there's more risk.

mglst:
I'll address your latest comment in a follow up PR

Issue #1073, opened 2026-09-13
Compress the image tree — 111 MB of unoptimized PNGs


mglst:
## Problem

`src/energetica/static/images/` is 111 MB across 70 tracked files, committed at full resolution. Essentially all of it is in git; there is no Git LFS and no `.gitattributes`.

The landing page is the worst of it. Seven PNGs in `landing_page/` are each **exactly 7,212,554 bytes**, which is not a coincidence — they are unoptimized exports from one pipeline. That folder alone is 53 MB, and the landing serves those banners to every first-time visitor.

Largest tracked blobs:

```
7212554  landing_page/resources_management.png
7212554  landing_page/renewable_intermittency.png
7212554  landing_page/investment_strategies.png
7212554  landing_page/dynamic_energy_markets2.png
7212554  landing_page/dynamic_energy_markets.png
7212554  landing_page/daily_knowledge_boost.png
7212554  landing_page/collective_action.png
6589758  storage_facilities/large_pumped_hydro.png
5597352  power_facilities/PV_solar.png
4243429  wiki/cloud_coverage.png
```

Per-folder counts: `wiki/` 16, `power_facilities/` 14, `technologies/` 12, `landing_page/` 11, `storage_facilities/` 6, `functional_facilities/` 4, `extraction_facilities/` 3, `icons/` 1.

## Why this matters more than the directory layout

This is user-facing page-weight on the landing page, which is the first thing a prospective player or educator loads. #1064 tidies where these files live; this makes the site faster. Of the two, this is the one players notice.

It also dissolves a secondary complaint. The tree is currently rsynced to every instance and a second time to the landing DocumentRoot. At 111 MB that duplication looks alarming; at a tenth of that it is unremarkable, and no longer worth designing around.

## What to build

Resize to the dimensions actually rendered, re-encode, and consider WebP with PNG fallback for the landing banners. Decide whether to rewrite history or simply commit the smaller files going forward — rewriting is the only way to shrink the clone, but it is disruptive and the repo is already this size.

## Out of scope

Moving the directory. That is #1064, and the two should not land in the same diff: a file move and a content rewrite want different reviewers and different verification.

## Context

Came out of grilling #1064.

Issue #1064, opened 2026-09-12
Move the app bundle out of the Python package


mglst:
> **Scope narrowed — see the comment below.** This issue originally covered the whole of `src/energetica/static/`. Three of its four bullets have since moved elsewhere. What remains is the app bundle and the service worker.

## Problem

`src/energetica/static/app/` and `src/energetica/static/service-worker.js` are build output from a different language, sitting inside a Python package. They are a vestige of the Flask/Jinja era: the `templates/` directory they were paired with is already gone, and nothing in the backend serves them — Apache does, straight off the rsynced tree.

Two of the three frontend bundles already do the right thing and only one doesn't:

```
vite.config.lobby.ts    → dist-lobby
vite.config.landing.ts  → dist-landing
vite.config.ts          → ../src/energetica/static/app     ← the odd one out
```

#1050 already excluded these paths from the wheel, so the distribution is clean. What remains wrong is the directory layout, not what ships.

**This is the change that empties `src/energetica/static/` and deletes it.** Every other occupant is accounted for elsewhere:

| | Destination | Where |
| --- | --- | --- |
| `static/data/` | `freeplay/data/` | #1049 step 2 |
| `static/images/` | `frontend/src/assets/`, bundled | #1078 |
| `icon_green.png` | `frontend/public/` | #1078 |
| `static/app/` | `frontend/dist-app/` | here |
| `static/service-worker.js` | `frontend/dist-app/` | here |

## Sequencing

**#1078 goes first.** It removes the image tree from this issue's scope entirely, so doing this one first would mean relocating that tree only to have #1078 delete the destination. #1078 needs no server-side work — it edits code and `deploy-landing.sh` — so the `Alias /static/images` line it orphans can be left dangling until this issue's vhost migration sweeps it up. One vhost pass instead of two.

#1049 step 2 can land in any order relative to this, but not concurrently: both touch `pyproject.toml`'s `package-data` glob.

## Constraint: the public URLs here are frozen

`/static/app/...`, `/service-worker.js` and `/manifest.json` keep their current URLs. Only filesystem paths change, so the blast radius is Apache `Alias` right-hand sides plus build config.

Two things make this non-negotiable:

[...]

PR #1083, opened 2026-09-14
refactor(images): bundle the image tree so a broken reference fails the build

### LINKED ISSUE #1078: Let Vite own the image assets, so a broken reference is a build error
## Problem

Nothing checks that an image reference points at a file that exists.

`src/energetica/static/images/` sits outside `frontend/`, so Vite never sees it. A path like `"/static/images/landing_page/hero.webp"` is an opaque string that happens to survive into the bundle. Rename the file, typo the path, or delete the image, and the build stays green — the failure surfaces as a broken image in someone's browser.

The frontend already does the right thing elsewhere. `frontend/src/assets/` holds three SVGs consumed as `import Logo from "@/assets/icon.svg?react"`, where Vite resolves the path at build time and a missing file fails the build. Only the image tree is exempt, and it holds the other ~34 references.

The worst of them are the ones built at runtime:

```ts
// frontend/src/components/facilities/facility-item.tsx
const imageUrl = `/static/images/${facilityType}_facilities/${facilityName}.webp`;
```

Thirty-nine facility and technology names reach this shape from the API across four components. Nothing connects them to the files on disk. During #1073 that correspondence had to be checked by hand, with a throwaway script comparing `enums.py` against the tree.

The cost of having no check is not hypothetical. #1073 also turned up five images nothing referenced — 9.5 MB before compression, including one 7.2 MB banner — sitting in the tree, rsynced to production on every deploy, for as long as nobody happened to look.

## What to build

**1. Move the image tree inside the bundle graph and reference it by import.**

Plain imports cover the landing and educators images. The dynamic lookups need `import.meta.glob`:

```ts
const facilityImages = import.meta.glob<string>(
    "@/assets/facilities/*/*.webp",
    { eager: true, import: "default", query: "?url" },
);
```

That is a compile-time record of path to hashed URL. Derive a union type from its keys and a wrong facility name becomes a `tsc` error rather than a 404. A hand-written file listing the joins is equally acceptable if it carries the same guarantee — the requirement is the guarantee, not the mechanism.

The MDX figures are the bulk of the work and the fiddliest part. `@mdx-js/rollup` currently runs `remarkGfm`, `remarkMath`, `rehypeSlug` and `rehypeKatex` — nothing that touches image nodes, so `alt` compiles to a literal string and gets no checking. Covering it means adding a remark plugin that rewrites image nodes into imports, and switching those paths to relative. Sixteen of the ~34 references live in six `.mdx` files.

**2. Keep `icon_green.png` in `frontend/public/`.**

It is the one file that genuinely needs a stable URL no bundler owns, because both its consumers sit outside Vite's graph:

- `frontend/public/manifest.json` is copied verbatim; Vite never parses its contents, so it cannot reference a hashed filename.
- `frontend/src/service-worker.ts` is built by `bun build`, a separate bundler. Written as an import it would resolve through bun's own pipeline and emit a second copy with a different hash. It also builds to a single `--outfile`, so there is no output directory for an asset to land in.

`icons/quiz.png` does **not** belong in this category and should become an ordinary import like everything else. It has one consumer, `frontend/src/routes/app/dashboard/quiz.tsx`, inside Vite's graph. At 3,485 bytes it also falls under Vite's default `assetsInlineLimit` of 4,096, so it would be inlined as a data URI and stop being a request at all.

## Why this is worth doing beyond the type safety

**Less gets shipped, not more.** Today both consumers receive the whole tree: `deploy-instance.sh` rsyncs 5.2 MB to the instance vhost and `deploy-landing.sh` rsyncs the same 5.2 MB into the landing DocumentRoot. 10.3 MB, most of it dead weight at its destination — the landing host receives 3.2 MB of facility art it has no route to serve. Under imports each bundle emits only what it uses: roughly 3.9 MB for the app, 1.9 MB for the landing, nothing for the lobby. About 5.8 MB total, with duplication confined to the 856 KB of wiki figures both genuinely share.

**Unreferenced images stop shipping by construction.** Unreferenced means unbundled means unsynced. The five dead files above could not have accumulated.

**Content hashing makes `Cache-Control: immutable` safe.** `apache-instance.conf` grants year-long immutable caching to `/static/app/assets/` only; the image tree gets no cache headers at all, so every repeat visitor revalidates all 40 images. A stable filename can never safely be marked immutable, because a changed image would be stuck in caches. Hashing is what removes that objection.

## Sequencing against #1064 and #1073

#1064 moves the tree to a top-level `static/` and freezes the public URLs, on the grounds that ETHZ owns `apache-ethz-proxy.conf`. This issue eventually *removes* `/static/images/` and its URLs, which makes the two partly redundant: if this lands first, #1064 has much less tree left to move. They should not be worked in parallel. Decide which goes first before starting either.

#1073 (compression) is already done and independent. It also added `scripts/guard-image-weight.ts`, which this change would supersede — Vite verifying references is a stronger guarantee than a CI check on the committed files. Delete the guard as part of this work.

## Out of scope

Generating `manifest.json` at build time, and moving the service worker from `bun build` to Vite. Both are reasonable and both were considered. The manifest becomes worth generating when it actually carries per-instance metadata, which `scripts/infra/apache-instance.conf:51` already anticipates; doing it now is machinery for one icon. Moving the service worker would still leave it special-cased on output name, scope and tsconfig, so the uniformity would be largely cosmetic.

mglst:
Closes #1078. Shrinks #1064.

> Stacked on #1080 — review that one first. This PR's own diff is the image tree move plus about twenty source files.

## The problem

Nothing checked that an image reference pointed at a file that existed. `src/energetica/static/images/` sat outside `frontend/`, so the bundler never saw it, and `"/static/images/landing_page/hero.webp"` was an opaque string that happened to survive into the bundle. Rename the file, typo the path, or delete the image, and the build stayed green — the failure surfaced as a broken image in someone's browser.

The worst of them were built at runtime, `` `/static/images/${facilityType}_facilities/${facilityName}.webp` ``, with thirty-nine facility and technology names reaching that shape from the API across four components and nothing connecting them to the files on disk.

## What this does

The tree moves to `frontend/src/assets/` and every image is reached through an import. Three mechanisms carry the guarantee.

**Facility and technology artwork** is looked up through `assetImages` in `lib/assets/asset-images.ts`, typed `Record<ProjectType, string>` off the generated API schema. The compiler checks the join in both directions: a project type the backend adds without artwork is a missing key, artwork for a type it no longer has is an excess property. The four components index that record instead of building a URL.

A hand-written map rather than `import.meta.glob`, deliberately. The glob types as `Record<string, string>`, which carries no key information at all — `tsc` would accept any string and, under `noUncheckedIndexedAccess`, hand back `string | undefined` for a runtime check. That is a weaker guarantee than the one the issue asked for. The issue allows either mechanism ("the requirement is the guarantee, not the mechanism").

**MDX figures** go through a new remark plugin, `frontend/remark-image-imports.ts`, which rewrites image sources into imports before the page compiles. `remark-mdx-images` handles only markdown `image` nodes, and eleven of the sixteen figures are JSX `` elements carrying `className`, so this one covers both syntaxes. Nine unit tests.

The figures use the `@/assets/…` alias rather than the relative paths the issue suggested, matching how the rest of the frontend imports — `eslint.config.js` bans `../*` imports repo-wide. The plugin accepts both, so the guarantee is identical either way. Relative paths stay supported because `alt` is the form an author reaches for first, and silently passing that through would be the exact failure this removes.

**`icon_green.png`** stays unhashed in `frontend/public/`. Both its consumers sit outside the module graph: `manifest.json` is copied verbatim, and `service-worker.ts` is built by `bun build`. Both now name it at `/static/app/icon_green.png`, which the existing alias already serves — no vhost change. `quiz.png` became an ordinary import and, at 3,485 bytes, falls under `assetsInlineLimit` and stops being a request at all.

## The guarantee, demonstrated

Typo an MDX figure path:

```
[vite:asset] Could not load .../src/assets/wiki/cloud_coverage_TYPO.webp
  (imported by src/content/wiki/time-and-weather.mdx)
```

Drop one entry from the facility registry:

```
error TS2741: Property 'coal_mine' is missing in type '{ steam_engine: string; ... }'
  but required in type 'Record<"mineral_extraction" | "laboratory" | ... , string>'.
```

## Less gets shipped

Each bundle emits only what it reaches:

| | before | after |
|---|---|---|
| app host | 5.2 MB | 3.9 MB |
| landing host | 5.2 MB | 1.5 MB |
| lobby | — | nothing |
| **total rsynced** | **10.3 MB** | **5.4 MB** |

`deploy-landing.sh` drops its second rsync of the whole tree. It also stops excluding `static/`, which is what lets `--delete` prune the old copy from hosts that still carry it.

The five images nothing referenced are deleted, including `icon.svg`, which was byte-identical to `frontend/src/assets/icon.svg`. Unreferenced now means unbundled means unshipped, so that class of file cannot accumulate again.

Hashed output lands in `/static/app/assets/`, already covered by the year-long `immutable` rule that a stable filename could never safely take.

## Two things to decide

**1. I did not delete `guard-image-weight.ts`, and I think the issue is wrong to ask for it.** The stated reason — "verifying references is a stronger guarantee than a CI check on the committed files" — covers references, but that is not what the guard checks. It checks that raster images are WebP and that no single file exceeds 400 KB. No bundler has an opinion on either. Delete it and someone can drop a full-resolution export into `frontend/src/assets/`, and it will be faithfully bundled, every megabyte of it — which is the regression #1073 existed to prevent.

It is retargeted at `frontend/src/assets/` and `frontend/public/` instead, passing over 68 files. Say the word and it goes: `git rm scripts/guard-image-weight.ts` plus the `package.json` script and the CI step.

**2. `bun run test` does not run in CI.** `frontend-checks` runs format, typecheck, lint and four guards, but no test step — so the nine plugin tests here, and the pre-existing `vite.shared.test.ts`, only ever run locally. One entry fixes it, but adding a CI job felt like a separate decision from this issue.

## Left alone on purpose

`Alias /static/images` and its `<Directory>` grant stay in `apache-instance.conf`, now pointing at a directory that no longer exists, with a comment saying so. Apache 404s it, for a path nothing requests any more. #1064 removes both lines in the same vhost migration that retargets the other aliases, so the two changes share one window instead of needing one each — per the ordering decided in the #1078 comment thread.

Out of scope and untouched: generating `manifest.json` at build time, and moving the service worker off `bun build`.

## Incidental

- `check_wiki_links.py` learns to read `src="…"` and to resolve a figure on disk. Without it every figure would have gone from a checked `200` to a silent `skip` the moment the URLs went away. `wiki-links.md` is tracked, so regenerating it also refreshed a few rows that were already stale before this branch.
- `optimize_images.py` is retargeted at the same two roots as the guard.
- The pre-commit eslint hook gains `--no-warn-ignored`. It names staged files explicitly, so touching the eslint-ignored `src/service-worker.ts` made eslint warn and `--max-warnings 0` fail the commit — a trap for any change that touches the worker, not just this one.

## Checks

`typecheck` (4 projects), `lint`, `format:check`, 97 unit tests, `build:all`, all 4 guards, 381 pytest, pyright clean, `check_wiki_links.py` 55/55.

<h2><a href="https://app.greptile.com/api/retrigger?id=64068429"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=1"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=1">Retrigger</picture></a>Confidence Score: 5/5</h2>

The PR appears safe to merge; no actionable regressions were identified in image resolution, serving, or deployment.

<h3>Summary</h3>

- Adds an exhaustive, API-typed facility and technology artwork registry.
- Converts Markdown and JSX wiki figures into imports through a shared remark plugin, with compilation tests.
- Keeps the PWA icon in public assets and updates both consumers.
- Removes the separate landing image deployment and retargets image maintenance tooling.
- No actionable regressions were identified.

<h3>Diagram</h3>

```mermaid
flowchart LR
    Components[React image imports] --> Vite[Vite asset resolution]
    Registry[ProjectType artwork registry] --> Components
    MDX[MDX figure paths] --> Remark[remarkImageImports]
    Remark --> Vite
    Vite --> App[App hashed assets]
    Vite --> Landing[Landing hashed assets]
    Public[Public PWA icon] --> AppRoot[App output root]
    App --> Apache[Apache static serving]
    AppRoot --> Apache
    Landing --> Deploy[Landing bundle rsync]
```

<sub>Reviews (1) · Last reviewed commit: ["refactor(images): let Vite own the image..."](https://github.com/felixvonsamson/energetica/commit/d6b4c12b4b6a87620d4f633c6e12e7ef56770394)</sub>


mglst:
@greptile-apps

felixvonsamson (approved):
cool

Issue #1051, opened 2026-09-12
Add an import-boundary contract in CI before the carve begins


mglst:
## Parent

Part of #1049 (Restructure the backend into layered packages).

## What to build

The dependency direction between packages becomes a declarative contract checked in CI — installed **before** anything is carved, so that every later ticket tightens one shared mechanism instead of inventing its own verification.

The contract initially encodes only what is already true: the lobby and the account/configuration layer must not pull in the persistent world's engine, routers, or real-time layer. It therefore passes on the unchanged tree. This is the prefactor that makes the rest of the sequence easy.

## Acceptance criteria

- [ ] An import-boundary linter is a declared dependency and runs in CI beside the existing linters
- [ ] A contract expresses the boundary that exists today and passes against the current tree
- [ ] A deliberately violating import is demonstrated to fail CI, then reverted
- [ ] The existing hand-written module-boundary test still passes and is **not** removed yet
- [ ] The full suite passes with no assertion changed

## Blocked by

- #1050

mglst:
I had a chat with Claude and I don't think this issue is really all that necessary. It's some CI stuff and it feels speculative. If down the line it needs to come back, we can revisit.

Issue #1049, opened 2026-09-12
Restructure the backend into layered packages, on a src/ layout, before the rest of Workshop Mode lands


mglst:
## Problem Statement

The backend lives in one package, `energetica`, that is three things at once: the persistent-world game, the shared platform every service depends on, and — since #993 — the beginning of Workshop Mode. A developer opening it cannot tell which of those a given module belongs to. 69 of its 113 modules reference `Player`; the rest are a mix of account storage, session signing, instance configuration, HTTP plumbing, and pure market maths, sitting side by side with no marker distinguishing them.

The boundary that matters is already real but invisible. The lobby service reaches into `energetica` for accounts, sessions, instance configuration, and error types, and `create_app` lazily imports the game graph specifically so that the lobby and the account layer never pull the persistent world in with them. That boundary is enforced today by a single test asserting four hand-written import pairs, whose own docstring calls the eventual fix "the follow-up identity-package extraction." Nothing else stops a new module from crossing it.

Workshop Mode makes this urgent rather than merely untidy. Workshop is a second Run format, and the decisions already taken mean it shares far less with the persistent world than the shared package name implies: it authors its own facility configuration rather than translating the persistent world's (#975), it cannot use the persistent world's player identity because that requires a `HexTile` it has no way to supply (#993), and it has no continuously-running clock at all. Every Workshop ticket therefore has to re-argue, in prose, which pieces transfer — #992's implementation-decisions section is largely that argument. Twenty-five such tickets remain — the Workshop Mode series runs to twenty-six, of which only the first is in flight. Each one merged against the current structure adds another module to a package whose name promises sharing it does not deliver.

Separately, the project is not installable. There is no `[project]` or `[build-system]` table; `energetica` and `lobby` resolve off the current working directory, which is why every systemd unit pins `WorkingDirectory`, every developer script changes directory first, and the OpenAPI generator manipulates `sys.path` by hand. Development and production import the same code by accident of where the process happens to start.

## Solution

Restructure the backend into named layers under a `src/` layout, with a one-directional dependency rule enforced by a lint contract rather than by convention, before the remaining Workshop Mode work lands.

`energetica` stays as the namespace — it is the product name and that part is not the problem. What changes is that every concern beneath it gets a name that says what it is:

```
src/energetica/
  kernel/     error types, session signing, hashing, version, formatting, action-log writing
  identity/   accounts store, instance configuration, server configuration, my-runs resolution
[...]

Issue #992, opened 2026-08-15
Workshop Mode — implementation spec


felixvonsamson:
> Implementation-ready spec synthesized from the closed [Workshop Mode wayfinder map](https://github.com/felixvonsamson/Energetica/issues/880) (#880) and its 25 closed child decision tickets. Nothing here is new design — see the map for provenance. Continued in the first comment below (GitHub's issue-body length limit).

# Workshop Mode

## Problem Statement

Energetica's normal game — the persistent world — is a slow-burn simulation. Players build power plants over real days and weeks, checking in for a few minutes at a time as their facilities finish construction and the market clears in the background. That's the right shape for a game people return to, but it's the wrong shape for teaching in a more limited time period. Felix wants to run electricity-market literacy sessions, in a single room, for two to four hours, with a facilitator standing at the front driving the room through a deliberately designed arc: build some capacity, set some prices, watch a market clear, get surprised by a shock, adjust, and see how decisions turn into money. None of that fits a simulation that runs for weeks with no one steering it.

Today there is no way to run that session inside Energetica. The free-play mode has no moderator concept, no fixed start-and-end session, no way to compress a "year" of market decisions into a few timed phases a room can sit through together, and no market participant standing in for the demand side when there's no real industry to consume the electricity. Retrofitting the free-play mode's mechanics — its per-player industry demand, its continuously-running clock, its player-initiated market-network joins, its multi-hour real-world construction times, its deep per-player tech tree — onto a two-hour classroom session would mean fighting the existing design at every turn, not reusing it.

Workshop Mode is the answer: a second, purpose-built Run format that reuses the pieces of the free-play mode that genuinely transfer (the uniform-price market clearing, the national demand curve's seasonal/intraday shape, the account and role system) and replaces everything else with mechanics sized for a moderated live session — a small number of Rounds, four Trading periods each, an always-on moderator console, and a facility catalog and event system authored specifically to teach market dynamics in a couple of hours rather than to model a persistent economy over weeks.

This document specs that format completely enough to build it. It does not build it.

## Solution

[...]

PR #1096, opened 2026-09-18
Carve the sim layer

### LINKED ISSUE #1054: Carve the sim layer
## Parent

Part of #1049 (Restructure the backend into layered packages).

## What to build

The shared simulation surface gets a name and a boundary: market clearing, the demand-shape function, and the facility-status vocabulary become a `sim` layer.

This layer is deliberately small — roughly 250 lines — because it is exactly, and only, what #992 names as reused by Workshop Mode. Its smallness is the point: once it exists, "does this transfer to Workshop?" is answered by looking at a directory rather than by re-arguing it in an issue.

Independent of the `kernel` and `identity` carves — these modules are pure leaves and import nothing.

## Acceptance criteria

- [ ] Market clearing, the demand-shape function, and the facility-status vocabulary live in a `sim` package beneath the namespace
- [ ] No module in `sim` imports from any other layer, and none references the persistent world's player or map concepts
- [ ] Every importer is updated; no re-exports remain at the old locations
- [ ] The layers contract names `sim` and passes
- [ ] The full suite passes with no assertion changed — the market-clearing tests differ only in import path

## Blocked by

- #1051

mglst:
Closes #1054. Part of #1049 (Restructure the backend into layered packages).

**Stacked on #1084.** The base is `feat/1064-move-app-bundle-out-of-python-package`, so this diff shows only the carve. Retarget to `main` once #1084 merges. The two changes are independent — they touch disjoint files, and #1049 notes they can land in either order.

## What this does

Market clearing, the demand-shape function, and the facility-status vocabulary move into `energetica.sim`:

| From | To |
| --- | --- |
| `energetica/market.py` | `energetica/sim/market.py` |
| `energetica/demand_shape.py` | `energetica/sim/demand_shape.py` |
| `energetica/types/facility_statuses.py` | `energetica/sim/facility_statuses.py` |

These three are the shared simulation surface: the pieces every mode of play simulates the same way, and exactly what the Workshop Mode spec (#992) names as reused from the persistent world.

The layer is a leaf. Its three modules import only the standard library, so nothing in it can reach a player, a map, or the game engine. It stays small on purpose — with it in place, "does this piece transfer to another mode?" is answered by whether the piece is in this directory rather than by re-arguing it in an issue.

`energetica/types/` held nothing but `facility_statuses.py`, so it is removed rather than left as an empty shell. Every importer points at the new path and no re-exports remain behind. Setuptools needed no change: `packages.find` globs `src/` and picks the new package up on its own.

This is a pure move. No assertion changed, `git diff -M` reports 100% similarity on all three files, and the market-clearing tests differ only in their import line.

## Acceptance criteria

- [x] Market clearing, the demand-shape function, and the facility-status vocabulary live in a `sim` package beneath the namespace
- [x] No module in `sim` imports from any other layer, and none references the persistent world's player or map concepts
- [x] Every importer is updated; no re-exports remain at the old locations
- [ ] **The layers contract names `sim` and passes** — see below
- [x] The full suite passes with no assertion changed

## The one criterion not met

The layers contract needs the Import Linter setup from #1051, which is deferred, so there is nothing yet to add a `sim` layer to. The property the contract would check does hold — every import statement in `src/energetica/sim/` is `__future__`, `math`, `dataclasses` or `typing` — but it is enforced by reading the imports, not by CI. Nothing stops a future `from energetica.database…` appearing inside `sim/` before #1051 lands.

I deliberately did not hand-roll a substitute test. #1049 asks for one mechanism rather than two, and a second hand-written boundary check is the thing #1051 exists to retire. **#1051 should name `sim` when it lands.**

## Checks

- `pytest`: 398 passed, 164 deselected (the `external_links` marker, excluded by default)
- `pyright`: 0 errors, 0 warnings
- `ruff check` and `ruff format --check`: clean

## Note for a reviewer

`MarketEntry.player_id` stays a bare `int`. It is not an import of the persistent world's `Player`, and the module docstring already records that clearing never reads it — that is what lets a caller inject a demand bid backed by no player at all. Changing it would be a behaviour change, which this ticket rules out.

<h2><a href="https://app.greptile.com/api/retrigger?id=66202085"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 5/5</h2>

The PR appears safe to merge because the moved implementations are unchanged, consumers use the new paths, and package discovery includes the new simulation package.

<h3>Summary</h3>

This PR carves reusable simulation primitives into a new leaf-level `energetica.sim` package without changing their implementations.
- Moves market clearing and demand-shape calculations with 100% similarity.
- Moves the shared facility-status vocabulary out of the removed `energetica.types` package.
- Updates production, player, schema, and unit-test imports to the new locations.
- Documents the new simulation package in the architecture overview.
- Package discovery includes the new package without additional configuration.

<h3>Diagram</h3>

```mermaid
flowchart LR
  Tick[production_update] --> Demand[sim.demand_shape]
  Tick --> Market[sim.market]
  Tick --> Status[sim.facility_statuses]
  Player[database.player] --> Status
  Schema[schemas.facilities] --> Status
  Tests[test_market_clearing] --> Market
  Demand --> Stdlib[Python standard library]
  Market --> Stdlib
  Status --> Stdlib
```

<sub>Reviews (1) · Last reviewed commit: ["refactor(backend): carve the sim layer"](https://github.com/felixvonsamson/energetica/commit/612daf170332c7511bbde50a3122b37083ac9c1c)</sub>


felixvonsamson:
So this also implies sharing the production logic ? what about all the fuction about fuel consumption, operational cost calculation, pollution calculation ? they should be shared too no ?

Issue #1097, opened 2026-09-20
Move market settlement arithmetic into sim


felixvonsamson:
## Parent

Part of #1049. Extends #1054 (the `sim` layer); see its "Extended scope" section for the reasoning.

## What to build

The rules that turn a market clearing into money move out of the `sell`/`buy` closures in `market_logic` and into a pure function in `sim`. The rules are the revenue conversion (MW × price × seconds ÷ 3600 ÷ 1e6), the 0.1 MW cutoff, the must-run offer convention, and the dump cost for unsold must-run power.

The new function takes a clearing and the seconds per tick and returns one result per market entry. The persistent world applies the results to `Player` and `new_values`.

The price floor also gets one name, `MIN_PRICE`, in place of the literal `-5` that appears five times in `production_update.py`. The `-6.0` in `market_optimum` is documented against it.

## Acceptance criteria

- [ ] The function lives in `energetica.sim`, takes plain values and returns plain data, and does not read `Player`, the engine or `new_values`
- [ ] `production_update.py` calls it, and the arithmetic is no longer written inline
- [ ] Before the move, a test pins the persistent world's current outputs for this piece. It passes unchanged after the move.
- [ ] The function has direct unit tests that construct no `Player` and no engine
- [ ] The full suite passes with no existing assertion changed
- [ ] `MIN_PRICE` replaces every literal `-5` price in `production_update.py`, and the `-6.0` in `market_optimum` is documented against it

## Blocked by

- #1084 landing on `main`. The `sim` package (#1096) is not on `main` until then.

## Not in this ticket

Workshop's own calls to this function. Those belong to the #992 sub-issues.

Issue #167, opened 2024-07-30
when market price is negative, dumping should be attributed to the player that is dumping to avoid other players to import to dump.


felixvonsamson:
![image](https://github.com/user-attachments/assets/b4edd12a-a296-42ba-befe-2f4dbdc219d9)
![image](https://github.com/user-attachments/assets/23444d59-4860-48da-b7ba-488a252f6e69)

mglst:
I'm not actually sure I agree!

PR #1103, opened 2026-09-20
Move market settlement arithmetic into sim

felixvonsamson:
Closes #1097. Part of #1049. Extends #1054.

**Stacked on #1084.** The base is `feat/1064-move-app-bundle-out-of-python-package`, which is the branch that contains the `sim` package (#1096). Retarget to `main` once #1084 merges.

## What this does

The rules that turn a market clearing into money move out of `market_logic` and into a pure function, `energetica.sim.settlement.settle_clearing`. Given a clearing and the length of a tick, it returns what each offer sold, what each demand bought, what must-run power was dumped, and the money involved. The rules are:

- revenue = MW × price × seconds ÷ 3600 ÷ 1e6
- a partly cleared entry below 0.1 MW counts as not having traded
- unsold must-run power (offered at the price floor) is dumped and billed at that price
- a demand that did not clear in full reports how much it received, so the caller can curtail it

`market_logic` now calls it and applies the results to players: money, generation and revenue records, curtailment, and chart data. That part stays in `production_update.py` because it reads `Player` and `new_values`.

The price floor is named `MIN_PRICE` in `sim/market.py`. It replaces the literal `-5` offer price in `production_update.py`, and the `-6.0` in `market_optimum` is now written as `MIN_PRICE - 1.0`.

## How it was checked

- Before the move, `tests/unit/test_market_settlement.py` pinned the current results for two small markets using real players: a partly cleared marginal offer with a curtailed bid, and unsold must-run power being dumped. It is the first commit and passes unchanged after the move.
- `tests/unit/test_settlement.py` tests `settle_clearing` directly, with no `Player` and no engine.
- `pytest`: 409 passed, with `tests/integration/test_server_boots.py` deselected. That test fails in my worktree only because its subprocess cannot import `energetica` from here, so please check it in CI.
- `pyright` reports 0 errors on the changed files. `ruff check` is clean, and the changed files are formatted (I used ruff 0.16.4 locally, not the pinned 0.12.10).

## Acceptance criteria

- [x] `settle_clearing` lives in `sim` and reads no `Player`, engine or `new_values`
- [x] `production_update.py` calls it and no longer repeats the arithmetic
- [x] `MIN_PRICE` replaces the literal `-5` price, and the `-6.0` is written against it
- [x] Current outputs are pinned before the move and unchanged after
- [x] Direct unit tests for the new function
- [x] Full suite passes with no existing assertion changed

## Note for a reviewer

The quantity below 0.1 MW is now written as `MIN_SETTLED_QUANTITY`. Its value is unchanged. `sim` still imports only the standard library.

<h2><a href="https://app.greptile.com/api/retrigger?id=66698255"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 5/5</h2>

The PR appears safe to merge, with settlement behavior preserved for reachable market inputs and no actionable defects identified.

<h3>Summary</h3>

This PR extracts electricity-market settlement arithmetic into a pure simulation function and leaves player-state application in the persistent-world update path.

- Adds typed settlement results for sales, purchases, dumping, revenue, costs, and partial demand service.
- Replaces duplicated settlement arithmetic in `market_logic` while preserving player balances, generation, curtailment, and chart updates.
- Introduces `MIN_PRICE` as the shared must-run offer floor and derives the market sentinel from it.
- Adds direct pure-function tests and player-backed characterization coverage.

<h3>Diagram</h3>

```mermaid
flowchart LR
    A[Supply offers and demand bids] --> B[clear_market]
    B --> C[MarketClearing]
    C --> D[settle_clearing]
    D --> E[Sales, purchases, and dumping]
    E --> F[market_logic]
    F --> G[Player money and revenues]
    F --> H[Generation and curtailment]
    F --> I[Market chart data]
```

<sub>Reviews (1) · Last reviewed commit: ["refactor(sim): move market settlement ar..."](https://github.com/felixvonsamson/energetica/commit/161440eee9f98d3f304bad8325eeffcf3eb30344)</sub>


src/energetica/sim/market.py
```diff
@@ -136,13 +142,13 @@ def market_optimum(offers: list[MarketEntry], demands: list[MarketEntry]) -> tup
...
     # For offers (sorted ascending): next step price is the price of the next row (or +inf for last)
-    # For demands (sorted descending): next step price is the price of the next row (or -6 for last)
+    # For demands (sorted descending): next step price is the price of the next row (or one below MIN_PRICE for last)
     events: list[tuple[float, bool, float]] = []
     for i, entry in enumerate(offers):
         next_price = offers[i + 1].price if i + 1 < len(offers) else math.inf
         events.append((entry.cumul_capacities, True, next_price))
     for i, entry in enumerate(demands):
-        next_price = demands[i + 1].price if i + 1 < len(demands) else -6.0
+        next_price = demands[i + 1].price if i + 1 < len(demands) else MIN_PRICE - 1.0
```
mglst: A value like `MIN_PRICE - 1.0` here seems pretty arbitrary. I suspect there's a good reason for it - @greptileai perhaps you can enlighten me - now I also appreciate that we might want to not touch this to preserve backwards compatibility but I still am curious as to why we're using this magic constant

mglst (approved):
This looks like a great change - I'm happy with the direction of abstraction

src/energetica/sim/market.py
mglst: @greptileai right - `-math.inf` does seem like a much more acceptable sentinel value. Would it lead to the same behaviour though if we just 'slotted it in'? Does the value of `-6.0` ever actually appear at runtime anywhere?

mglst: @greptileai then would you recommend making the switch to `-math.inf`?

src/energetica/sim/settlement.py
```diff
@@ -0,0 +1,113 @@
...
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+
+from energetica.sim.market import MIN_PRICE, MarketClearing
+
+#: Below this many MW a partly cleared entry is treated as not having traded at all.
+MIN_SETTLED_QUANTITY = 0.1
```
felixvonsamson, line 20: is that really MW ? if yes that rounds like a large treshhold. @Greptile-ai

src/energetica/sim/settlement.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+def energy_value(quantity: float, price: float, seconds_per_tick: float) -> float:
+    """Return the money for ``quantity`` MW at ``price`` (per MWh) sustained for one tick, in millions."""
```
felixvonsamson: Why in millions ?

felixvonsamson: I thought the quantitiy unit was W. That would explain the /1_000_000 here and the large treshhold above

src/energetica/sim/settlement.py
```diff
@@ -0,0 +1,113 @@
...
+
+@dataclass(frozen=True, slots=True)
+class SaleSettlement:
+    """What one supply offer sold, and what was dumped if it was must-run power that did not clear."""
+
+    player_id: int
+    facility: str
+    quantity: float  # MW sold at the market price; 0 when nothing traded
+    revenue: float  # money earned for ``quantity``; negative when the market price is negative
+    counts_as_generation: bool  # False for must-run offers, whose output the caller already recorded
```
felixvonsamson, line 40: the name is not great. "flexible_generation" would be better

src/energetica/sim/settlement.py
```diff
@@ -0,0 +1,113 @@
...
+
+@dataclass(frozen=True, slots=True)
+class PurchaseSettlement:
+    """What one demand bid bought, and how much of it was left unserved."""
+
+    player_id: int
+    facility: str
+    quantity: float  # MW bought at the market price; 0 when nothing traded
+    cost: float  # money owed for ``quantity``; negative when the market price is negative
+    served: float | None  # MW the bid received when it was not fully served; None when it was fully served
```
felixvonsamson: isn't quantity and server the same value when not 100% served. I would have expected to store either total quantity and quantity served or quantity and unserved energy.

src/energetica/sim/settlement.py
```diff
@@ -0,0 +1,113 @@
...
+    purchases: list[PurchaseSettlement]
+
+
+def settle_clearing(clearing: MarketClearing, seconds_per_tick: float) -> Settlement:
+    """Work out what each entry of ``clearing`` sold or bought, and what it costs or earns.
+
+    Offers arrive in merit order. Offers that cleared in full sell their whole capacity. The first offer
+    that did not clear in full sells what it can, and nothing after it sells, except that must-run offers
+    (priced at :data:`MIN_PRICE`, which sort first) that did not clear dump the power they could not sell.
+    Every demand that did not clear in full reports how much of it was served, so the caller can curtail it.
```
felixvonsamson, line 70: Wild idea I have that might simplify the logic. instead of setting the min price to -5 and force dumping if not cleared, what about setting must-run offers at -inf and have an demand block at -5 with infinite demand with the label dumping.
Not sure this is better. What do you  think @mglst ?

src/energetica/sim/settlement.py
mglst: @greptileai

src/energetica/sim/settlement.py
```diff
@@ -0,0 +1,113 @@
...
+
+@dataclass(frozen=True, slots=True)
+class Settlement:
+    """Every trade and dump of one clearing, in the order the caller should apply them."""
+
+    sales: list[SaleSettlement]
+    purchases: list[PurchaseSettlement]
+
+
+def settle_clearing(clearing: MarketClearing, seconds_per_tick: float) -> Settlement:
```
felixvonsamson, line 64: I'm womdering if it would not make more sense to merge that with the market clearing function since I have the feeling we duplicate the structure and the market clearing output has to run through this function anyway. @claude @mglst

src/energetica/sim/settlement.py
```diff
@@ -0,0 +1,113 @@
...
+    Offers arrive in merit order. Offers that cleared in full sell their whole capacity. The first offer
+    that did not clear in full sells what it can, and nothing after it sells, except that must-run offers
+    (priced at :data:`MIN_PRICE`, which sort first) that did not clear dump the power they could not sell.
+    Every demand that did not clear in full reports how much of it was served, so the caller can curtail it.
+    """
+    price = clearing.price
+    quantity = clearing.quantity
+
+    sales: list[SaleSettlement] = []
+    for fill in clearing.offers:
```
felixvonsamson, line 76: fill is a strange variable name

src/energetica/sim/settlement.py
mglst: I think the value in this, in having a special value that `served` can take is that we can take different code branches, for partial and for full satisfaction, without having to resort to checking equality with `served == quantity` which can be tricky / error prone because of floating point arithmetic

src/energetica/sim/settlement.py
mglst: Yes I'm totally on board. That makes a lot more sense.

mglst: I have another 'wild' idea I'm cooking with Claude

src/energetica/sim/settlement.py
mglst: Yeah idk if this is related by I don't like the shape of the functions that 'take in' a market and 'return' a 'new' market. That's a smell to me.

mglst:
@felixvonsamson See #167 - you say:

> when market price is negative, dumping should be attributed to the player who is dumping to avoid other players to import to dump.

Claude says that, right now:

> player dumps is decided by dict insertion order among offers tied at -5, so the attribution is arbitrary [...] Felix's -inf + dumping-block design doesn't lose per-player attribution — it stops pretending there was any

mglst:
Keeping this PR to what it says on the tin — a behaviour-preserving move of the settlement arithmetic into the sim layer — and moving the rest into #1111.

The review turned up four changes that belong together rather than separately, so they are written up as one spec for us to slice: dumping as a demand order (@felixvonsamson's idea above), retiring `counts_as_generation`, splitting the order book from the chart accumulator, and freezing the order. Each of them changes what the characterization tests in `test_market_settlement.py` assert, and those tests exist precisely to prove this PR changes nothing, so none of them can ride along here.

Two things from the review that are worth pulling out.

**Units: @felixvonsamson was right, the quantity is in W, not MW.** `config/assets.py` gives power in watts, and the chart schema documents capacity, cumulative capacity and cleared volume as watts with prices in $/MWh. So in `energy_value` the division by a million is a watts-to-megawatts conversion, not a scaling of money into millions: 820,000 W at $50/MWh for an hour is $41. Two consequences for this PR, both comment-only, and they are the only other change landing here:

- Every `# MW` comment in `settlement.py` is wrong, as is "in millions" in the `energy_value` docstring.
- `MIN_SETTLED_QUANTITY = 0.1` is 0.1 W, so the instinct inverts: it is not a large threshold, it is a near-zero epsilon.

There are three more comments in `production_update.py` that say `# MW tolerance` on a value in watts, including one ten lines below a docstring that correctly says watts. Those are pre-existing, but they are only comments, so they are corrected here too.

**The two uses of negative infinity collide.** This PR already took the suggestion to make the last demand step's sentinel `-math.inf`. If must-run supply orders are also priced at `-inf`, the crossover test `price_d < price_o` becomes `-inf < -inf`, which is false, so `market_optimum` never finds the crossing and falls through to `raise ValueError("No market optimum found.")`. Only one side can use infinity, and under the new design it has to be the supply side, so the demand sentinel needs to go back to a finite value below the must-run price. This is flagged as the easiest detail to miss in #1111.

Also, while in there: the adjacent comment still says "one below `MIN_PRICE`" while the code now says `-math.inf`. Corrected here.

---

On dumping attribution, for the record, since it decides #167. Running the real clearing code:

1. Dumping only ever happens when the clearing price is exactly the floor. Whenever a must-run order is the marginal unit the next offer price is still the floor, so the price is the floor. If demand exceeds must-run supply, all of it clears and nothing dumps.
2. Selling and dumping are therefore worth the same. An order of capacity `C` that sells `s` earns `-5s` and pays `5(C - s)`, which is `-5C` whatever `s` is.
3. The split is decided by insertion order. Swap two equal must-run `place_ask` calls and the two facilities trade places on the charts.

So the dumping demand order does not lose per-player attribution. There was never any to lose, which is why dumped power currently shows up as an *import* for the dumping player in the screenshots on #167.

felixvonsamson:
Thank you very much @mglst for the retargetting and clearing of this PR

Issue #1101, opened 2026-09-20
Move the operating-cost formula into sim


felixvonsamson:
## Parent

Part of #1049. Extends #1054 (the `sim` layer); see its "Extended scope" section for the reasoning.

## What to build

The rule that splits operating cost into a fixed share and a variable share scaled by utilisation moves into `sim`. The fixed share is 100% for renewables and storage, 50% for nuclear, and 20% for other controllable facilities and extraction.

The function takes the maximum cost for a period and the utilisation, and returns the cost owed. The caller chooses the period: the persistent world passes the per-tick cost, and Workshop will pass its per-Round maximum.

## Acceptance criteria

- [ ] The function lives in `energetica.sim`, takes plain values and returns plain data, and does not read `Player`, the engine or `new_values`
- [ ] `production_update.py` calls it, and the arithmetic is no longer written inline
- [ ] Before the move, a test pins the persistent world's current outputs for this piece. It passes unchanged after the move.
- [ ] The function has direct unit tests that construct no `Player` and no engine
- [ ] The full suite passes with no existing assertion changed

## Blocked by

- #1084 landing on `main`. The `sim` package (#1096) is not on `main` until then.

## Not in this ticket

Workshop's own calls to this function. Those belong to the #992 sub-issues.

PR #1107, opened 2026-09-20
Move the operating-cost formula into sim

felixvonsamson:
Closes #1101. Part of #1049. Extends #1054.

**Stacked on #1084.** The base is `feat/1064-move-app-bundle-out-of-python-package`, which contains the `sim` package (#1096). Retarget to `main` once #1084 merges. This PR is independent of the other extraction PRs.

## What this does

Every facility has a maximum operating (O&M) cost for a period. Part of it is fixed and owed however little the facility runs, and the rest scales with how much it ran. That rule moves out of `resources_and_pollution` into `sim/operating_cost.py`:

- `fixed_cost_share(facility, scales_with_use)`: 100% for renewable and storage facilities, 50% for the two nuclear reactors, 20% for other controllable facilities and for extraction facilities
- `operating_cost(maximum_cost, fixed_share, utilisation)`: the cost owed

The function does not know the length of the period. The persistent world passes the cost per tick. A mode that states costs per round, such as Workshop, passes its per-round maximum.

`resources_and_pollution` still works out the utilisation (generation over power, or extraction demand over power use) and applies the cost to the player. Those read the player's state, so they stay with the caller.

## How it was checked

- `tests/unit/test_operating_cost.py` is the first commit. It pins the split for a wind turbine, a battery, a coal burner, a nuclear reactor and a coal mine with a real player, including an idle facility and the total charged to the player. It passes unchanged after the move.
- `tests/unit/test_operating_cost_formula.py` tests the new functions directly. One of its tests checks that enum members are recognised as nuclear, because the persistent world passes enum members and the shared code matches on plain names.
- I ran the old and new `resources_and_pollution` side by side over 3,000 random ticks. Operating costs, emissions and the money charged were bit-for-bit identical. That script is not committed.
- `pytest`: 411 passed, with `tests/integration/test_server_boots.py` deselected (it can't import `energetica` from my worktree; please check it in CI).
- `ruff check` is clean and `pyright` reports 0 errors on the changed files.

<h2><a href="https://app.greptile.com/api/retrigger?id=66704912"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 5/5</h2>

The PR appears safe to merge with no actionable correctness, security, or quality issues identified.

<h3>Summary</h3>

Extracts the operating-cost calculation from `production_update` into a player-agnostic simulation module while preserving the existing charging behavior.
- Adds reusable fixed-cost-share and utilization-based cost functions.
- Keeps player-state utilization calculation and balance updates in `resources_and_pollution`.
- Adds direct formula tests and player-level characterization coverage.

<sub>Reviews (1) · Last reviewed commit: ["refactor(sim): move the operating-cost f..."](https://github.com/felixvonsamson/energetica/commit/f186ae1e4f498b6c4fbf022a4f3d5feee5255980)</sub>


mglst:
I think ... this ... I'm a bit skeptical of this.
I would think that these computations more belong to whatever class embodies facilities. And we discussed that these are indeed different between freeplay and workshop. There could be a common base perhaps. I've not actually looked at the cade changes in this PR at all yet - but just reading the title and description, I am cautious.

src/energetica/sim/operating_cost.py
```diff
@@ -0,0 +1,36 @@
...
+how much it ran. The caller chooses the period: the persistent world passes the cost per tick, and a mode
+that states its costs per round passes the cost per round.
+"""
+
+from __future__ import annotations
+
+#: Renewable and storage facilities do not save anything by running less, so all of their cost is fixed.
+FIXED_SHARE_ALWAYS_OWED = 1.0
+#: The fixed share of nuclear reactors, which are costly to run below capacity.
+FIXED_SHARE_NUCLEAR = 0.5
```
mglst, line 14: yeah so like these kinds of constants - they 'belong' to facilities. It feels wrong to have them as just hardcoded, screaming case, constants in the code base.

src/energetica/sim/operating_cost.py
```diff
@@ -0,0 +1,36 @@
...
+#: The fixed share of every other controllable facility and of extraction facilities.
+FIXED_SHARE_DEFAULT = 0.2
+
+NUCLEAR_FACILITIES = frozenset({"nuclear_reactor", "nuclear_reactor_gen4"})
```
mglst, line 18: Same hardcoded complaint here. I'd think that these would be _flags_ / booleans in the config for each facility. I actually think this very function / functionality (answering "is this facility nuclear? / which facilities are nuclear?") is already answered by one of the `enum` files, where the fuel types are defined.

mglst (changes requested):
Yeah idk about this one. The code changes are very minimal when excluding test - good I suppose - or at least reassuring - but they don't feel right, they don't feel like the right design.

